PR #25, opened 2024-12-25
Optimization Pass

Hawxy:
A couple minor items I've had sitting around for a while.

- Adjust SourceWriter to be backed by a string builder pool and get rid of intermediary string allocations if possible
- Move to generated regex where possible
- Sealed generated code and added `GeneratedCodeAttribute` to classes for better tooling support.
- Fast-path a few spots.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -21,7 +21,7 @@
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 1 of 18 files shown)


jeremydmiller:
@Hawxy I'm getting this right now, but with a manual merge. One bit, and awesome sauce for these!

PR #83, opened 2025-06-21
Fall back to System Console if Spectre Console fails

LodewijkSioen:
Ran into [this bug](https://github.com/spectreconsole/spectre.console/issues/1495) in Spectre.Console. Took me a while to figure out since the exception was only thrown in my pipeline.

I think it makes sense to fall back to `System.Console` if Spectre fails. That way you always get the actual exception in the logs and not the exception Spectre throws.


LodewijkSioen:
The test that fails seems unrelated to the code in the PR...

PR #98, opened 2025-08-07
Fix for aggregation runner calling StoreProjection

micahosborne:
Not sure why it calls StoreProjection.  This always leads to a not implemented exception in marten.  

This is a reference to the bug test created here.  This pull request also has more information.
https://github.com/JasperFx/marten/pull/3898


jeremydmiller:
The "fix" is something different. `StoreProjection` is necessary to side step potential versioning issues and to version the projected documents correctly. Thank you though!

PR #112, opened 2025-10-09
Allow injecting unresolved nullable dependencies as null

YandyZaldivar:
Generic proposal for https://github.com/JasperFx/wolverine/issues/1738
Requires testing


YandyZaldivar:
@jeremydmiller sounds good targeting Wolverine 5.

I think it feels interesting explicitly opting for the dependency by making it nullable and it won't introduce a breaking change. If a dependency is required and it must be resolved, then there is no reason to make it nullable. We can also ask for another opinion just to validate this idea.

Other than that, I haven't tested the implementation, so we need to make sure it does the job before releasing.

PR #138, opened 2026-01-21
Allow registration of custom ValueTypeInfo

JurJean:
The default convention for value types doesn't work for us so we would like to be able to inject our own ValueTypeInfo.

### DIFF
--- a/src/JasperFx/Core/Reflection/ValueTypeInfo.cs
+++ b/src/JasperFx/Core/Reflection/ValueTypeInfo.cs
@@ -36,7 +36,7 @@ public class ValueTypeInfo
         if (builder != null)
         {
             valueType = new ValueTypeInfo(type, valueProperty.PropertyType, valueProperty, builder);
-            _valueTypes = _valueTypes.AddOrUpdate(type, valueType);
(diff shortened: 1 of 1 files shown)


JurJean:
Nothing special about our value types, the assumption that value types only and always have a `Value` property is just too simplistic.

We often have the 'HasValue' property for example, and I don't understand why someone couldn't add more properties to it however they see fit.

So we needed a way to explicitly point to the `Value` property.

Sadly the `Value` property seems to be hardcoded on other locations in the codebase, so this ValueTypeInfo.Register only fixes a part of the problem. There doesn't seem to be a way to nicely define a method (like Format(), ToString() for reading, Parse() for writing).

I'm sure there are reasons for it (backwards compatibility) but everything would be a lot easier if the serialized json value would be used instead, since then we just have to deal with these simple types only.

jeremydmiller:
@JurJean Don't comment on issues that are already closed if you're hoping for a response. I'd almost never had noticed this come in.

You're in "I take pull requests" territory here.

JurJean:
@jeremydmiller You asked _me_ a question.

PR #237, opened 2026-05-12
IJasperFxAggregateGrouper.Group: take IReadOnlyList<IEvent> (#201)

### LINKED ISSUE #201: IAggregateGrouper.Group: change IEnumerable<IEvent> to IReadOnlyList<IEvent>
## Summary

The `IAggregateGrouper<TId>.Group` method currently accepts `IEnumerable<IEvent>`:

```csharp
Task Group(TQuerySession session, IEnumerable<IEvent> events, IEventGrouping<TId> grouping);
```

Implementations that need to inspect the same set of events more than once — for example, first to collect grouping keys and then to resolve related document IDs — are forced to either enumerate twice or materialise the sequence themselves with a `.ToList()` call. Static analysis tools (Rider, ReSharper, Roslyn analysers) correctly flag this as a **possible multiple enumeration** warning, because `IEnumerable<IEvent>` gives no guarantee that re-iteration is safe or cheap.

## Suggested change

Change the parameter type to `IReadOnlyList<IEvent>` (or at minimum `IList<IEvent>`):

```csharp
Task Group(TQuerySession session, IReadOnlyList<IEvent> events, IEventGrouping<TId> grouping);
```

`IReadOnlyList<T>` communicates that:
- the collection is already materialised (safe to enumerate multiple times),
- callers can rely on `Count` and indexed access without allocating,
- implementors do not need a defensive `.ToList()` at the top of every `Group` method.

## Why this matters

A realistic `Group` implementation often needs two or more passes:

1. A first pass over all events to partition them by type and collect aggregate IDs.
2. One or more subsequent passes (or LINQ projections over the same collection) to resolve related document IDs from the database.

With `IEnumerable<IEvent>` the compiler and analyser cannot tell whether this is safe. With `IReadOnlyList<IEvent>` it is unambiguous and no defensive copy is needed.

## Breaking change?

Changing the interface signature is a breaking change for existing implementors. Given that Marten itself supplies the concrete `events` argument (always from a materialised in-memory list), a migration path could be:

- Introduce `IReadOnlyList<IEvent>` in a minor version with the old `IEnumerable<IEvent>` overload marked `[Obsolete]`.
- Remove the deprecated overload in the next major version.

Alternatively, if semver policy permits, a straight breaking change in a minor version is acceptable given that implementors only need to update the parameter type in their own class — no logic change required.

jeremydmiller:
## Summary
- Promote the \`events\` parameter on \`IJasperFxAggregateGrouper<TId, TQuerySession>.Group\` from \`IEnumerable<IEvent>\` to \`IReadOnlyList<IEvent>\`.
- Mirror the change through the internal \`LambdaAggregateGrouper\`, the \`EventSlicer.CustomGrouping(Func<...>)\` overload, and the same overload on \`JasperFxMultiStreamProjectionBase\`.

Closes #201.

## Why
Implementations of \`IJasperFxAggregateGrouper\` frequently need two or more passes over the same event batch — partition by type first, then resolve related document IDs from the database. \`IEnumerable<IEvent>\` gave no guarantee re-iteration was safe or cheap, so static analysers (Rider, ReSharper, Roslyn) flagged it as possible-multiple-enumeration. \`IReadOnlyList<IEvent>\` makes \`Count\` + indexed access first-class.

The call site (\`EventSlicer.SliceAsync\`) already received the events as \`IReadOnlyList<IEvent>\` — the materialised list always existed; the interface was just hiding it.

## Breaking change
Yes — \`IJasperFxAggregateGrouper\` implementors need to update one parameter type in their \`Group\` method (no logic change). JasperFx 2.0 is a major bump, so a straight break is acceptable per the issue's \"Breaking change?\" discussion.

## Test plan
- [x] \`dotnet build src/JasperFx.Events/JasperFx.Events.csproj\` — clean (only pre-existing warnings)
- [x] \`dotnet test src/EventTests/EventTests.csproj --framework net9.0\` — 255/255 pass

### DIFF
--- a/src/JasperFx.Events/Aggregation/JasperFxMultiStreamProjectionBase.cs
+++ b/src/JasperFx.Events/Aggregation/JasperFxMultiStreamProjectionBase.cs
@@ -213,7 +213,7 @@ public abstract class JasperFxMultiStreamProjectionBase<TDoc, TId, TOperations, 
     /// </summary>
     /// <param name="func"></param>
     /// <exception cref="InvalidOperationException"></exception>
-    public void CustomGrouping(Func<TQuerySession, IEnumerable<IEvent>, IEventGrouping<TId>, Task> func)
+    public void CustomGrouping(Func<TQuerySession, IReadOnlyList<IEvent>, IEventGrouping<TId>, Task> func)
     {
         if (_customSlicer != null)
         {
--- a/src/JasperFx.Events/Grouping/IEventSlicer.cs
+++ b/src/JasperFx.Events/Grouping/IEventSlicer.cs
@@ -69,25 +69,35 @@ public interface IEventSlicer<TDoc, TId, TQuerySession> where TId : notnull
 public interface IJasperFxAggregateGrouper<out TId, in TQuerySession>
 {
     /// <summary>
-    ///     Apply custom grouping rules to apply events to one or many aggregates
+    ///     Apply custom grouping rules to apply events to one or many aggregates.
+    ///     <para>
+    ///     #201: the parameter is <see cref="IReadOnlyList{T}"/> rather than
+    ///     <see cref="IEnumerable{T}"/> — implementations frequently need two or
+    ///     more passes over the same events (e.g. partition by type, then resolve
+    ///     related document IDs from the database), and <see cref="IEnumerable{T}"/>
+    ///     gave no guarantee re-iteration was safe or cheap, forcing every
+    ///     implementor to either eat the multi-enum warning or do a defensive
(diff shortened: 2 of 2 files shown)


jeremydmiller:
Note: external contributor @erdtsieck opened #202 first (April 30) covering the same change. We can pick either — closing this in favour of theirs would respect their first-mover; merging this would land slightly richer doc comments around the rationale (the cross-pass-with-database-lookup motivation from the issue body).

jeremydmiller:
Used @erdtsieck's change

PR #359, opened 2026-05-22
Fix source generator evolver and hint-name failures

erdtsieck:
## What changed

- Treat a generated partial projection as covering its aggregate type so the snapshot-registration pipeline does not emit a second standalone evolver for the same aggregate.
- Reuse the aggregate constructor helper for generated Evolve/EvolveAsync fallback paths.
- Avoid emitting required-member object initializers for aggregate types without a public parameterless constructor.
- Normalize aggregate types discovered from snapshot/ref-aggregate call sites so nullable reference annotations do not leak into generated evolvers.
- Sanitize source-generator hint names before calling `AddSource()` so Roslyn-invalid characters such as `?` cannot crash generation.
- Add regression tests for required-member primary-constructor aggregates with conventional projection Create/Apply methods and nullable `[ReadAggregate]` aggregate parameters discovered via `IRefersToAggregate` attributes.

## Why

Two source-generator paths could break Marten 9 upgrade scenarios:

- A partial projection could generate the correct dispatcher while a later source-generator pipeline also emitted a fallback evolver for the same aggregate. That fallback did not know about the projection Create method and could generate construction code that fails for aggregates using required members and primary constructors.
- Aggregate types discovered from nullable parameters such as `[ReadAggregate] EigenPrestatie? prestatie` could feed `?` into generated hint names, causing Roslyn to reject `AddSource()` and abort the whole generator pass.

Addresses JasperFx/marten#4542 and JasperFx/marten#4543.

## Validation

- `dotnet test .\src\JasperFx.Events.SourceGenerator.Tests\JasperFx.Events.SourceGenerator.Tests.csproj --no-restore`

### DIFF
--- a/src/JasperFx.Events.SourceGenerator.Tests/AggregateEvolverGeneratorTests.cs
+++ b/src/JasperFx.Events.SourceGenerator.Tests/AggregateEvolverGeneratorTests.cs
@@ -665,6 +665,65 @@ public class Registration
         allGenerated.ShouldContain("[assembly: global::JasperFx.Events.Aggregation.GeneratedEvolver(typeof(global::Test.Counter)");
     }
 
+    [Fact]
+    public void nullable_aggregate_parameter_attribute_does_not_crash_hint_name_generation()
+    {
+        var source = @"
+#nullable enable
+using System;
+using JasperFx.Events;
+using JasperFx.Events.Aggregation;
+
+namespace Test;
+
+public class ReadAggregateAttribute(string memberName) : Attribute, IRefersToAggregate;
+
+public record EigenPrestatie
+{
+    public required string Id { get; init; }
+    public required string Prestatiecode { get; init; }
+
+    public static EigenPrestatie Create(EigenPrestatieAangemaakt e) => new()
+    {
+        Id = e.PrestatieId,
+        Prestatiecode = e.Prestatiecode
+    };
+}
+
+public class EigenPrestatieAangemaakt
+{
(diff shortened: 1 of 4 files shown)


jeremydmiller:
Verified this locally and it's the complete fix for marten#4542 (+ marten#4543). 👍

- Ran the suite on this branch: **18/18 green**, including the reporter-faithful `partial_projection_for_required_member_aggregate_suppresses_snapshot_fallback` and the `nullable_aggregate_parameter_attribute_does_not_crash_hint_name_generation` (CS8785) case.
- I'd independently built a focused construction-only fix (#362) with 3 extra regression tests (record→`GetUninitializedObject`, plain-class→`default!` initializer, reporter-shape compiles). This branch passes all 3 of those too → **21/21** when combined. So #359 is a strict superset; I've closed #362 in favor of it.
- Also reproduced the underlying construction bug end-to-end against Marten `release/9.0.0`: a partial `SingleStreamProjection` over a required-member primary-ctor record fails to build (CS7036) with the released generator.

Two small things I'd double-check before merge (both look fine on read):
1. `MarkSeen` only suppresses the standalone evolver when `Mode == PartialProjection && Methods.Count > 0` — so a `Snapshot<T>` with *no* covering partial projection still gets its evolver. ✅
2. The `default!` guard now keys off `TypeKind.Struct || InstanceConstructors.Any(public parameterless)`.

Happy to push my 3 construction-focused tests onto this branch if you'd like the extra coverage.

PR #362, opened 2026-05-22
Fix SG evolver construction for required-member aggregates (marten#4542)

jeremydmiller:
## What changed

The source-generated evolver's Apply-only / null-snapshot branch synthesized `new T { Required = default! }` for **any** aggregate with `required` members. For a primary-constructor record — e.g. `record DiagnostiekActiviteit(string Id) { required Guid? SubContractorId; ... }` — that object initializer needs a parameterless constructor the record doesn't have, so the generated `*.Evolver.g.cs` failed to compile with **CS7036** (or **CS9035** when the id was supplied positionally).

This guards the `default!` object initializer on the type actually having an **accessible public parameterless constructor** (or being a struct). When there isn't one (a primary-ctor record), it falls back to `RuntimeHelpers.GetUninitializedObject(typeof(T))`, which allocates without invoking a ctor or enforcing required members; `Apply` then populates them. `default!` is now only ever emitted when there's a public no-arg ctor for the initializer to call.

## Why

Addresses [JasperFx/marten#4542](https://github.com/JasperFx/marten/issues/4542). A required-member record aggregate projected by a partial `SingleStreamProjection` with conventional `Create`/`Apply` produced uncompilable generated code.

## Validation

- `dotnet test src/JasperFx.Events.SourceGenerator.Tests` — green (added regression tests: record-with-`Create`/`Apply` compiles; record `Apply`-only synthesizes via `GetUninitializedObject`, never `default!`; required-member **plain class** still synthesizes via the `default!` object initializer — Marten's `sample_external-account-link` pattern).
- **End-to-end against Marten `release/9.0.0`:** a partial `SingleStreamProjection` over a required-member primary-ctor record **fails to build** with the released generator (CS7036) and both **builds and runs correctly** with this fix (event stream → `Create` → `Apply` → loaded document has every required member set).

## Relationship to #359

This converges on the **same construction rule** as #359 (emit `default!` only with a public parameterless ctor; otherwise `GetUninitializedObject`). #359 is broader — it also fixes marten#4543 (the nullable hint-name `AddSource` crash) and removes the redundant standalone-evolver emission when a partial projection already covers the aggregate.

This PR is a focused, regression-tested fix for the marten#4542 **construction** bug. Maintainer's call whether to land this and bring #359's marten#4543 + duplicate-evolver fixes separately, or land #359 as the superset.

### DIFF
--- a/src/JasperFx.Events.SourceGenerator.Tests/AggregateEvolverGeneratorTests.cs
+++ b/src/JasperFx.Events.SourceGenerator.Tests/AggregateEvolverGeneratorTests.cs
@@ -733,4 +733,162 @@ public class UpdatedEvent { }
         evolver.ShouldContain("case global::Test.CreatedEvent");
         evolver.ShouldContain("case global::Test.UpdatedEvent");
     }
+
+    [Fact]
+    public void required_members_with_primary_ctor_and_conventional_projection_compiles()
+    {
+        // marten#4542: a record aggregate with a primary constructor for the id
+        // PLUS required members, projected by a partial SingleStreamProjection
+        // with conventional Create/Apply. The projection's Create fully
+        // initialises the aggregate, so the generator must NOT emit a standalone
+        // fallback evolver whose `new T(data)` ignores the required members
+        // (CS9035). The record's primary ctor (string Id) is the id-construction
+        // parameter, not an event-shaped implicit Create handler.
+        var source = @"
+using System;
+using JasperFx.Events;
+using JasperFx.Events.Aggregation;
+using JasperFx.Events.Projections;
+
+namespace Test;
+
+public record DiagnostiekActiviteit(string Id)
+{
+    public required Guid? SubContractorId { get; set; }
+    public required string Aanlevercode { get; set; }
+    public required int Prestatiecodelijst { get; set; }
+}
+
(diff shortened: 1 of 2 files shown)


jeremydmiller:
Closing in favor of #359, which is the more complete fix.

I verified empirically that this focused PR resolves only a *narrower* slice of marten#4542 (the partial-projection construction expression). When I added #359's own test cases to this branch, both failed:

- **`partial_projection_for_required_member_aggregate_suppresses_snapshot_fallback`** — the faithful reporter scenario (partial projection + `opts.Snapshot<T>()` + the record's event ctor) still emits a redundant standalone evolver whose `new T(data)` event-ctor path fails with **CS9035**. This PR's construction-expression change doesn't touch the event-ctor path or suppress the duplicate evolver — #359's `MarkSeen` change does.
- **`nullable_aggregate_parameter_attribute_does_not_crash_hint_name_generation`** — **CS8785** hint-name crash (marten#4543), which this PR doesn't address.

Conversely, #359 passes its own 18 tests **and** the 3 construction-focused tests from this PR (21/21), so it's a strict superset and converges on the same construction rule (emit `default!` only with an accessible public parameterless ctor; otherwise `GetUninitializedObject`).

Thanks @erdtsieck — #359 is the one to land. Happy to contribute the 3 extra construction-focused regression tests from here if useful.

PR #482, opened 2026-06-30
Per-tenant agent distribution + database-affine assignment surface (wolverine#3280, marten#4806)

erdtsieck:
Foundational piece of the per-tenant agent distribution work for sharded + tenant-partitioned event stores. Addresses JasperFx/wolverine#3280 and carries the store-agnostic surface for JasperFx/marten#4806.

**Problem.** Under node-distributed daemons (Wolverine-managed event-subscription distribution), agents are distributed per (shard × database). With `MultiTenantedWithShardedDatabases` + `UseTenantPartitionedEvents`, multiple tenants are co-located in one shard database, each drawing its own event sequence — a single store-global agent (and high-water mark) per database cannot track them, so a lagging tenant's appends fall below the shared mark and are skipped.

**What's here** (default interface members — zero impact on stores that don't opt in):

- `IEventStore.DistributesAgentsPerTenant` (default `false`): signals that a node-distributed daemon must fan out one agent per (shard, tenant). Store-global, single-database per-tenant partitioning, and database-per-tenant behavior is unchanged.
- `JasperFxAsyncDaemon.StartAgentAsync(string)` detects a tenant-bearing identity (`ShardName.TryParse` → non-null `TenantId`), resolves the base shard, activates the tenant in the high-water coordinator, and starts a per-tenant agent that advances against its own mark and persists its own `<proj>:All:<tenant>` progression row — polling the per-tenant high water right after start.
- `IEventStore.GroupAgentAssignmentsByDatabase` (default `false`): opt-in signal for **database-affine assignment** (marten#4806) — a distribution-aware host keeps all of a shard database's per-tenant agents together on one node, so connection pools scale with the number of shard databases instead of nodes × databases (which otherwise exhausts a shared server's `max_connections`).
- Defaults covered by `AgentDistributionDefaultTests` (same bare-store convention as `AllDatabasesDefaultTests`).

**Companion PRs (merge order):** this first → Marten #4799 (surfaces both signals + per-tenant high-water and bulk-import work) → Wolverine #3281 (per-tenant fan-out + affine distribution).

### DIFF
--- a/src/EventTests/AgentDistributionDefaultTests.cs
+++ b/src/EventTests/AgentDistributionDefaultTests.cs
@@ -16,10 +16,9 @@ public class AgentDistributionDefaultTests
 {
     // A bare IEventStore that overrides none of the agent-distribution members, so the assertions below
     // exercise the default interface implementations. A store that does not opt in must behave exactly as
-    // before: one agent per shard×database (DistributesAgentsPerTenant false), even per-agent spreading
-    // (GroupAgentAssignmentsByDatabase false), and no fan-out bound (MaxNodesPerDatabaseForAgents 1).
-    // See jasperfx/wolverine#3280 and JasperFx/marten#4806. Everything else throws — those members are
-    // never invoked here.
+    // before: one agent per shard×database (DistributesAgentsPerTenant false) and even per-agent spreading
+    // (GroupAgentAssignmentsByDatabase false). See jasperfx/wolverine#3280 and JasperFx/marten#4806.
+    // Everything else throws — those members are never invoked here.
     private sealed class BareEventStore : IEventStore
     {
         public Task<EventStoreUsage?> TryCreateUsage(CancellationToken token) => throw new NotImplementedException();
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Thank you @erdtsieck — this design and nearly all of the code carry forward as-is. Superseded by #487, which contains every commit from this branch with your authorship intact, plus three review-driven revisions on top:

1. `GroupAgentAssignmentsByDatabase` is dropped from `IEventStore` — database-affine grouping will be always-on in Wolverine's distribution for any store reporting a multi-database `DatabaseCardinality`, rather than a per-store opt-in contract.
2. The per-tenant `StartAgentAsync` branch now primes the tenant's high-water ceiling *before* starting the agent (a `SubscribeFromPresent` subscription otherwise resolved "present" against ceiling 0 and replayed the tenant's full history).
3. Hardenings: exact-identity precedence over tenant parsing, loud failure for tenant identities on non-partitioned stores, and 7 new daemon-level behavioral tests.

Tracked under the per-tenant partitioning epic #486.

PR #511, opened 2026-07-12
Do not restart an already-started host in RunCommand. Closes #510

### LINKED ISSUE #510: RunCommand double-starts a host already started under AutoStartHost, re-running every IHostedService.StartAsync
### Describe the bug

When `JasperFxEnvironment.AutoStartHost` is set — which `WebApplicationFactory`-based test harnesses (e.g. Alba) do — the command executor starts the host **twice**. `buildExecutor` eager-starts the pre-built host, then `RunCommand.Execute` calls `StartAsync` on the same instance. `IHost.StartAsync` is not re-entrant, so every `IHostedService.StartAsync` runs a second time, after `ApplicationStarted` has already fired. Hosted services with non-idempotent startup are corrupted, and tests that assert on start-time state become flaky.

Same instance across both starts:

- `buildExecutor` — `src/JasperFx/CommandLineHostingExtensions.cs:158`:
  ```csharp
  if (JasperFxEnvironment.AutoStartHost && source is PreBuiltHostBuilder b)
  {
      b.Host.Start();   // start #1 — fires ApplicationStarted
  }
  ```
- `RunCommand.Execute` — `src/JasperFx/CommandLine/Commands/RunCommand.cs`:
  ```csharp
  using var host = input.BuildHost();   // PreBuiltHostBuilder.Build() returns the same started host
  ...
  await host.StartAsync();              // start #2 — re-runs every IHostedService.StartAsync
  ```

### To Reproduce

A complete console app. Verified against **JasperFx 2.27.0** — `dotnet run` prints `IHostedService.StartAsync invocations: 2` / `BUG — host started twice` and exits `1`.

`Repro.csproj`:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <Nullable>enable</Nullable>
    <ImplicitUsings>enable</ImplicitUsings>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="JasperFx" Version="2.27.0" />
  </ItemGroup>
</Project>
```

`Program.cs`:

```csharp
using JasperFx;
using JasperFx.CommandLine;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;

// WebApplicationFactory-based test harnesses (for example Alba) set this so JasperFx
// starts the IHost while Program.Main runs, giving the test a live server.
JasperFxEnvironment.AutoStartHost = true;

// Signalled by the hosted service's *second* StartAsync — the redundant start this repro detects.
var redundantStart = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
var counter = new StartCounter(() => redundantStart.TrySetResult());

var host = Host.CreateDefaultBuilder()
    .ConfigureServices(services => services.AddSingleton<IHostedService>(counter))
    .Build();

// The run command blocks until shutdown, so drive it on a background thread.
var runTask = Task.Run(() => host.RunJasperFxCommands([]));

// Do not shut down until the redundant start has happened (or a short bound elapses),
// otherwise a stopping host would mask it.
await Task.WhenAny(redundantStart.Task, Task.Delay(TimeSpan.FromSeconds(2)));

host.Services.GetRequiredService<IHostApplicationLifetime>().StopApplication();
await runTask;

Console.WriteLine($"IHostedService.StartAsync invocations: {counter.Starts}");
Console.WriteLine(counter.Starts == 1 ? "OK — host started once" : "BUG — host started twice");
Environment.Exit(counter.Starts == 1 ? 0 : 1);

internal sealed class StartCounter(Action onSecondStart) : IHostedService
{
    private int _starts;
    public int Starts => _starts;

    public Task StartAsync(CancellationToken cancellationToken)
    {
        if (Interlocked.Increment(ref _starts) == 2)
        {
            onSecondStart();
        }

        return Task.CompletedTask;
    }

    public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
}
```

### Expected behavior

Each `IHostedService.StartAsync` runs once. A host already started (e.g. by `buildExecutor`'s eager start) should not be started again by `RunCommand`.

### Additional context

We hit this as a ~40% flaky test in a Wolverine.HTTP + Alba app: a hosted service registered at index 0 was observed running a second time *after* `ApplicationStarted` — structurally impossible in a single start, since an index-0 service completes before `NotifyStarted`.

A one-line guard in `RunCommand.Execute` fixes it — skip `StartAsync` when the host's lifetime reports it already started. Reading the actual host state (rather than the `AutoStartHost` flag) keeps it correct for a caller that set the flag but was not eager-started (a plain `IHostBuilder`, not a `PreBuiltHostBuilder`):

```csharp
if (lifetime is null || !lifetime.ApplicationStarted.IsCancellationRequested)
{
    await host.StartAsync();
}
```

Implemented with a regression test in #511.

uniquelau:
Fixes #510: under `JasperFxEnvironment.AutoStartHost` (WebApplicationFactory testing), `buildExecutor` eager-starts the pre-built host and `RunCommand.Execute` then calls `StartAsync` on the same instance. `IHost.StartAsync` is not re-entrant, so every `IHostedService.StartAsync` runs a second time after `ApplicationStarted` has fired.

## The change

One guard in `RunCommand.Execute` — skip the start when the host already started:

```csharp
if (lifetime is null || !lifetime.ApplicationStarted.IsCancellationRequested)
{
    await host.StartAsync();
}
```

**Why read `ApplicationStarted` and not the `AutoStartHost` flag?** `buildExecutor` only eager-starts when `AutoStartHost && source is PreBuiltHostBuilder`. A caller that set the flag but ran through the plain `IHostBuilder` overload is *not* eager-started, and `RunCommand` should start it. Reading the lifetime asks the actual host instance, so it is correct on every path; the flag is not.

- Preserves the normal CLI path: without an eager start, `ApplicationStarted` is not yet signalled, so the guard starts the host exactly as before.
- Leaves `buildExecutor`'s eager start untouched — other commands (describe, resources) rely on a live host, and command resolution happens after `buildExecutor`, so the guard belongs at the redundant start in `RunCommand`.

## Testing

New `run_command_compliance` regression test drives the real `RunJasperFxCommands` path under `AutoStartHost` and asserts a single start. It gates on the redundant second start (a `TaskCompletionSource` signalled by the counter's second `StartAsync`), so it fails fast when the bug is present and a bounded wait proves the negative when it is fixed — no stopping the host early, which would mask the redundant start. Fails without the guard (`Starts == 2`), passes with it; full `CommandLineTests` suite green (295).

### DIFF
--- a/src/CommandLineTests/run_command_compliance.cs
+++ b/src/CommandLineTests/run_command_compliance.cs
@@ -1,4 +1,7 @@
+using JasperFx;
+using JasperFx.CommandLine;
 using JasperFx.CommandLine.Commands;
+using JasperFx.Core;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Shouldly;
@@ -7,6 +10,62 @@ namespace CommandLineTests;
 
 public class run_command_compliance
 {
+    [Fact]
+    public async Task does_not_restart_a_host_already_started_by_auto_start()
+    {
+        // JasperFxEnvironment.AutoStartHost (WebApplicationFactory testing) makes the command
+        // executor eager-start the pre-built host before the run command executes. The run
+        // command must not start it a second time — IHost.StartAsync is not re-entrant and would
+        // re-run every IHostedService.StartAsync.
+        JasperFxEnvironment.AutoStartHost = true;
+        try
+        {
+            // Signalled by the counter's second StartAsync — i.e. the redundant start we guard
+            // against. It fires promptly when the bug is present and never when it is fixed.
+            var redundantStart = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+            var counter = new StartCounter(onSecondStart: () => redundantStart.TrySetResult());
+
+            var host = Host.CreateDefaultBuilder()
+                .ConfigureServices(services => services.AddSingleton<IHostedService>(counter))
+                .Build();
+
+            // Run on a background thread: with the guard in place the run command skips its only
+            // await, so its shutdown Wait() blocks synchronously — a direct call would deadlock.
+            var runTask = Task.Run(() => host.RunJasperFxCommands([]));
+
+            // Do not shut down until any redundant start has happened, or a stopping host would
+            // mask it. A correct run never signals, so bound the wait to prove the negative.
+            await Task.WhenAny(redundantStart.Task, Task.Delay(2.Seconds()));
+
+            host.Services.GetRequiredService<IHostApplicationLifetime>().StopApplication();
+            await runTask.TimeoutAfterAsync(5000);
+
+            counter.Starts.ShouldBe(1);
+        }
+        finally
+        {
+            JasperFxEnvironment.AutoStartHost = false;
+        }
+    }
+
+    public class StartCounter(Action? onSecondStart = null) : IHostedService
+    {
+        private int _starts;
+        public int Starts => _starts;
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (Interlocked.Increment(ref _starts) == 2)
+            {
+                onSecondStart?.Invoke();
+            }
+
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+    }
+
     [Theory]
     [InlineData(null)]
     [InlineData("")]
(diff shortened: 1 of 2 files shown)


jeremydmiller (approved):
Reviewed in depth: host-state-based detection (reading ApplicationStarted off the actual instance) is the right call over trusting the AutoStartHost flag — it correctly handles all four AutoStartHost × builder-source combinations, including the plain-IHostBuilder-with-AutoStartHost case where a flag check would wrongly skip, and it fixes the hand-started-host path for free. Shutdown symmetry verified (one start, one stop, either way). The TCS-gated regression test failing on revert and passing on the fix is exactly the right shape. Verified locally: 13/13 on the new test class, 295/295 on the full CommandLineTests suite. Approved the workflow run — merging when CI completes.

PR #543, opened 2026-07-21
GH-3521 (Wolverine): warn on divergent process-pinned application assembly

jeremydmiller:
Ports the [wolverine#3521](https://github.com/JasperFx/wolverine/issues/3521) diagnostic into JasperFx itself. The Wolverine side (asks 1 & 2) already shipped in [wolverine#3530](https://github.com/JasperFx/wolverine/pull/3530); that PR flagged the **`JasperFxOptions` twin** as the follow-up, since `JasperFxOptions.RememberedApplicationAssembly` is the *dominant* process-wide static (Wolverine reads `jasperfx.ApplicationAssembly` first).

## The trap

The application assembly JasperFx scans for code generation and type discovery is cached in a **process-wide static** pinned by whichever host starts **first** in the process. In a normal single-host app that's harmless. In a **test process that stands up multiple Critter Stack hosts across different assemblies**, a later implicit host silently inherits the first host's assembly — so types defined in a *different* assembly vanish for the whole run. It's **order-dependent**: passes in isolation, fails in full-suite runs, flips between builds as recompilation reshuffles test order.

## The fix (diagnosability only)

Wolverine already detects this for **Wolverine** hosts. This makes JasperFx detect it for **any** JasperFx host (Marten, etc.):

1. `AddJasperFx` captures the registering assembly (`DetermineCallingAssembly()`) **while the caller's frame is still live**, and stashes it on the options instance as `RegistrationCallingAssembly`.
2. `establishApplicationAssembly` compares the **adopted** (process-pinned) assembly against it and, on divergence, records a human-readable `ApplicationAssemblyReuseWarning`.

**Detection-only** (per discussion on the Wolverine side): JasperFx exposes `ApplicationAssemblyReuseWarning` as a **public property**; consumers that have a logger at startup surface it. JasperFx is a library with no always-on startup hosted service, so it intentionally does **not** emit here — Wolverine already emits its own (richer) message, and other consumers can read the property. No double-firing.

An **explicit** `ApplicationAssembly` / `SetApplicationProject` stays **silent** (never reaches `establishApplicationAssembly`), and the **first** host never diverges from itself.

### AOT note

`DetermineCallingAssembly()` now runs on every `AddJasperFx` (previously only when `RememberedApplicationAssembly == null`). Multi-host test processes are never AOT-published; a single AOT host that sets the application assembly explicitly routes through the `ApplicationAssembly != null` short-circuit, and under NativeAOT the stack frames carry no method metadata so the call falls back to `Assembly.GetEntryAssembly()` without throwing and its result is unused. The `[UnconditionalSuppressMessage]` justification is updated accordingly.

## Tests

`JasperFxOptionsTests` (CoreTests), with `RememberedApplicationAssembly` static save/restore:
- `warns_when_adopted_assembly_diverges_from_where_the_host_registered` — red-verified (fails with the divergence check disabled).
- `does_not_warn_when_the_adopted_assembly_matches_where_the_host_registered`
- `does_not_warn_when_the_registration_assembly_could_not_be_resolved`
- `does_not_warn_when_the_application_assembly_is_set_explicitly`
- `a_normal_single_assembly_host_does_not_warn` — real-host false-positive guard; also pins that the registration assembly is captured as the test assembly (not `JasperFx`).

## Not addressed

- No new always-on hosted service in JasperFx (would be a behavior addition to every Critter Stack host); emission stays with consumers.
- Wolverine needs no change — it already covers Wolverine hosts. A future Marten follow-up could surface `ApplicationAssemblyReuseWarning` from its own startup path.

### DIFF
--- a/src/CoreTests/JasperFxOptionsTests.cs
+++ b/src/CoreTests/JasperFxOptionsTests.cs
@@ -219,4 +219,122 @@ public class JasperFxOptionsTests
             DynamicCodeBuilder.WithinCodegenCommand = originalValue;
         }
     }
+
+    // GH-3521: the application assembly is a process-wide value pinned by whichever host starts FIRST in the
+    // process. A later host that adopts it while it was registered from a DIFFERENT assembly silently scans
+    // the wrong assembly. These tests pin the behavior of the divergence detection surfaced on JasperFxOptions.
+
+    [Fact]
+    public void warns_when_adopted_assembly_diverges_from_where_the_host_registered()
+    {
+        var original = JasperFxOptions.RememberedApplicationAssembly;
+        try
+        {
+            // Simulate an earlier host in the process having pinned a different assembly...
+            var pinned = typeof(JasperFxOptions).Assembly;
+            JasperFxOptions.RememberedApplicationAssembly = pinned;
+
+            var options = new JasperFxOptions
+            {
+                // ...while THIS host was registered from the test assembly.
+                RegistrationCallingAssembly = GetType().Assembly
+            };
+
+            options.ReadHostEnvironment(new StubHostEnvironment());
+
+            options.ApplicationAssembly.ShouldBe(pinned);
+            options.ApplicationAssemblyReuseWarning.ShouldNotBeNull();
+            options.ApplicationAssemblyReuseWarning.ShouldContain(pinned.GetName().Name!);
+            options.ApplicationAssemblyReuseWarning.ShouldContain(GetType().Assembly.GetName().Name!);
+        }
+        finally
+        {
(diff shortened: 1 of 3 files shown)


jeremydmiller:
Downstream emit-side follow-ups filed (detection-only means consumers surface `ApplicationAssemblyReuseWarning`):
- Marten: JasperFx/marten#4996 (log from `MartenActivator`)
- Polecat: JasperFx/polecat#345 (needs an always-on emit point — `PolecatActivator` is conditional)
- Wolverine: JasperFx/wolverine#3576 (reconcile with Wolverine's existing #3530 detection; avoid double-warning)

All gated on a JasperFx release containing this PR + pin bumps.

PR #561, opened 2026-07-23
Slice ID support for RaiseSideEffects

lofbacke:
Added RaiseSideEffects overload to support injecting strongly-typed slide ID into parameters while maintaining backwards-compatibility.

### DIFF
--- a/src/JasperFx.Events/Aggregation/JasperFxAggregationProjectionBase.cs
+++ b/src/JasperFx.Events/Aggregation/JasperFxAggregationProjectionBase.cs
@@ -472,7 +472,20 @@ public abstract partial class JasperFxAggregationProjectionBase<TDoc, TId, TOper
     {
         return new ValueTask();
     }
-    
+
+    /// <summary>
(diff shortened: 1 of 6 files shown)


src/JasperFx.Events/Aggregation/JasperFxAggregationProjectionBase.cs
```diff
@@ -472,7 +472,20 @@ public virtual ValueTask RaiseSideEffects(TOperations operations, IEventSlice<TD
     {
         return new ValueTask();
     }
-    
+
+    /// <summary>
+    /// Potentially raise "side effects" during projection processing to either emit additional events,
+    /// or publish messages. The identity of the current slice is supplied as <paramref name="id"/>.
```
RikNedap, line 478: It is not clear (to me) what "the identity of the current slice" means. Is it the identity of TDoc?

RikNedap (commented):
I don't what the requirements are around here for automated tests, but there aren't any in this PR.

This new feature should probably also be in the [documentation](https://martendb.io/events/projections/side-effects.html)

Issue #600, opened 2026-08-02
DetermineCallingAssembly adopts the test-runner assembly, so type discovery scans a runner instead of the test assembly


jeremydmiller:
`JasperFxOptions.DetermineCallingAssembly()` can resolve the application assembly to the **test runner** (`xunit.v3.core`) rather than the test assembly. Every consumer that uses `JasperFxOptions.ApplicationAssembly` for type discovery then scans an assembly containing none of the suite's types.

This was the root cause of JasperFx/wolverine#3776 (fix: JasperFx/wolverine#3777). Wolverine has been defended at its own boundary, but the defect is here and Marten / Polecat hosts have the same exposure.

## The code

`JasperFxOptions.cs:243`:

```csharp
if (assemblyName.StartsWith("System") || assemblyName.StartsWith("Microsoft") || assemblyName.StartsWith("ReSharperTestRunner"))
{
    continue;
}

return assembly;
```

The walk takes the first frame outside JasperFx that isn't `System*`/`Microsoft*`/`ReSharperTestRunner*`. Under an **async test fixture** the frames between JasperFx and the test class belong to the runner, not the test assembly — so `xunit.v3.core` satisfies the filter and is returned. The `ReSharperTestRunner` entry shows this class of problem was already known; xUnit was simply never added.

The short-circuit above it does not save you: `HasReferenceToJasperFxTool(Assembly.GetEntryAssembly())` returns `False` for an ordinary test project (no referenced assembly carries `[JasperFxTool]`), so the stack walk always runs. I verified this directly — `False` at both versions tested, with zero reference-load failures.

## Measured impact

Instrumented CI runs of one Wolverine test shard, logging the adopted assembly per host, at a single fixed commit:

| | adopted assembly | outcome |
|---|---|---|
| before fix | `xunit.v3.core` on **79 of 82** hosts, 0–15 handler chains | 43 of 86 tests failed |
| after fix | `PolecatTests` on **82 of 82**, 141 chains | 86 passed, 0 retries |

The symptom is entirely downstream and gives no hint of the cause — in Wolverine's case `IndeterminateRoutesException: Could not determine any valid subscribers or local handlers`. For Marten the equivalent would be a document/projection type simply not being discovered.

[...]

PR #604, opened 2026-08-02
fix: attribute registration to the app, not the extension that registered for it (#601)

### LINKED ISSUE #601: DetermineCallingAssembly returns intermediate Critter Stack assemblies, making the GH-3521 divergence warning fire on healthy hosts
Follow-up to #600 — same function, different failure mode. Where #600 is about the walk landing on a *test runner*, this is about it landing on an *intermediate Critter Stack assembly*, which makes the GH-3521 divergence warning unreliable as a diagnostic.

## What I measured

Instrumenting Wolverine's twin of this logic on a **fully green** local run of one test shard: `RegistrationCallingAssembly` resolved to **`Wolverine.SqlServer` on 81 of 82 hosts**, not to the test assembly that actually called `UseWolverine`. Because the adopted application assembly was correctly `PolecatTests`, `checkForDivergentApplicationAssembly` compared `Wolverine.SqlServer` against `PolecatTests`, found them different, and raised the warning on all 81 — on a run where nothing was wrong.

The measurement is from Wolverine's copy of the walk. I have not instrumented JasperFx's directly, but the two are structurally identical (`DetermineCallingAssembly` → `RegistrationCallingAssembly` set in `AddJasperFx` → compared in `checkForDivergentApplicationAssembly` at `JasperFxOptions.cs:171`), and the cause applies equally: when a Critter Stack extension registers JasperFx on behalf of the application, the first non-`System*`/`Microsoft*` frame outside JasperFx belongs to **the extension**, not the app.

## Why it matters

The warning exists to make a silent first-host-wins pin observable. Firing it on healthy hosts inverts that value:

- It trains readers to ignore it.
- Its **absence** stops being evidence. On JasperFx/wolverine#3776 I explicitly recorded "zero occurrences of the warning in the failing logs" as grounds for ruling out an application-assembly problem. That was exactly the bug, and the reasoning was void — partly because of this noise, partly because of a separate hole on the other branch.

## Possible directions

1. Skip Critter Stack framework assemblies the same way runners are skipped, so `RegistrationCallingAssembly` means what its name says. Needs care — it must not skip a legitimate application assembly that happens to be named for a product.
2. Capture the registering assembly at the outermost public entry point rather than re-deriving it from a stack walk deep in the registration path.
3. If neither is practical, only warn when the adopted assembly is one the walk is known to resolve unreliably, rather than on every name mismatch.

I do not have a strong recommendation between these; the current behaviour is the actionable part.

https://claude.ai/code/session_013eR4GL278688VhyhrGcttJ

jeremydmiller:
Fixes #601.

> **Stacked on #602.** Both change the same walk, so this PR targets `gh600/skip-test-runner-assemblies` rather than `main`. Review #602 first; this diff is the second commit only. Retarget to `main` once #602 merges.

## The defect

Follow-up to #600 — same function, different failure mode. Where #600 was about the walk landing on a *test runner*, this is about it landing on an *intermediate Critter Stack assembly*.

`UseWolverine()` and `AddMarten()` both call `services.AddJasperFx()` from inside their own assembly (`Wolverine/HostBuilderExtensions.cs:109`, `Marten/MartenServiceCollectionExtensions.cs:210,331`), so the first frame outside JasperFx belongs to the **extension**, not the application.

## Reproduced, not taken on faith

The issue measured Wolverine's twin of this logic and noted JasperFx's own walk had not been instrumented directly. It has now. A stand-in extension assembly that calls `AddJasperFx()` on the app's behalf, in a real host:

```
RegistrationCallingAssembly = Wolverine.StackWalkStandIn   <- the extension
ApplicationAssembly         = CoreTests                    <- correct
Warning                     = FIRED                        <- on a host where nothing is wrong
```

That inverts the warning's whole value. It trains readers to ignore it, and its **absence** stops being evidence — on JasperFx/wolverine#3776, "zero occurrences of the warning in the failing logs" was explicitly recorded as grounds for ruling out an application-assembly problem, and that was exactly the bug.

## It is more than warning noise

The same stack walk seeds the process-wide `RememberedApplicationAssembly` in `AddJasperFx`. Where `IHostEnvironment.ApplicationName` is empty, `establishApplicationAssembly` falls through to that pin — so the **extension assembly became the assembly type discovery scans**. Measured on a bare `ServiceCollection`:

```
RegistrationCallingAssembly=Widgets1 | Remembered=Widgets1 | ApplicationAssembly=Widgets1 | Warning=none
```

Note `Warning=none` there: registered and adopted agree, because both are wrong the same way. The divergence check is silent in exactly the case where it is most needed.

## The fix

Skip Critter Stack framework assemblies in the walk, alongside the runners:

```csharp
if (assemblyName.StartsWith("System") || assemblyName.StartsWith("Microsoft") ||
    IsTestRunnerAssembly(assemblyName) || IsCritterStackAssembly(assemblyName))
{
    continue;
}
```

Two guards against the risk the issue flags — *"it must not skip a legitimate application assembly that happens to be named for a product"*:

1. **Exact name or dotted prefix**, never a bare `StartsWith`. `Wolverine` and `Wolverine.SqlServer` match; `WolverineDemo`, `MartenPlayground` and `JasperFxSamples` do not.
2. **Never a `*.Tests` / `*.Testing` assembly.** `Wolverine.RabbitMQ.Tests` and `Marten.Testing` ARE the application as far as discovery is concerned — their handlers and documents are the types under test. Skipping those would reintroduce #600 from the other direction.

List: `JasperFx` · `Wolverine` · `WolverineFx` · `Marten` · `Weasel` · `Polecat` · `CritterWatch` · `Oakton`.

## Tests

`src/ExtensionStandIn` is a project whose only purpose is its assembly name — `Wolverine.StackWalkStandIn` — with an extension method that calls `AddJasperFx()` on the caller's behalf. Same trick as `TestRunnerStandIn` in #602.

- `registration_is_attributed_to_the_app_not_the_extension_that_registered_for_it` — real host; asserts `RegistrationCallingAssembly` is the test assembly **and** that the divergence warning stays quiet.
- `an_extension_registration_does_not_pin_the_extension_as_the_application_assembly` — the no-`ApplicationName` path; asserts neither `ApplicationAssembly` nor the process-wide pin becomes the extension.
- Predicate theories covering the framework names, the look-alike application names, and the in-repo test suites.

Both behaviour tests **fail** without the fix and pass with it. Full `CoreTests` green (539 passed, 1 skipped).

## Not done here

The issue's option 2 — capturing the registering assembly at the outermost public entry point instead of re-deriving it from a stack walk — is the structurally better answer and is unaffected by this change. This is the version that needs no downstream work in Wolverine or Marten to take effect.

### DIFF
--- a/jasperfx.slnx
+++ b/jasperfx.slnx
@@ -41,6 +41,7 @@
     <Project Path="src/Widgets4/Widgets4.csproj" />
     <Project Path="src/Widgets5/Widgets5.csproj" />
     <Project Path="src/TestRunnerStandIn/TestRunnerStandIn.csproj" />
+    <Project Path="src/ExtensionStandIn/ExtensionStandIn.csproj" />
   </Folder>
   <Project Path="build/Build.csproj">
     <Build Project="false" />
--- a/src/CoreTests/CoreTests.csproj
+++ b/src/CoreTests/CoreTests.csproj
@@ -28,6 +28,7 @@
         <ProjectReference Include="..\FSharpTypes\FSharpTypes.fsproj" />
         <ProjectReference Include="..\Widgets5\Widgets5.csproj" />
         <ProjectReference Include="..\TestRunnerStandIn\TestRunnerStandIn.csproj" />
+        <ProjectReference Include="..\ExtensionStandIn\ExtensionStandIn.csproj" />
     </ItemGroup>
 
 </Project>
--- a/src/CoreTests/JasperFxOptionsTests.cs
+++ b/src/CoreTests/JasperFxOptionsTests.cs
@@ -1,3 +1,4 @@
+using ExtensionStandIn;
 using JasperFx;
 using JasperFx.CodeGeneration;
 using JasperFx.CommandLine.Descriptions;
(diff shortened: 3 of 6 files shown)


jeremydmiller:
Auto-closed by GitHub when its stacked base branch (`gh600/skip-test-runner-assemblies`) was deleted on merge of #602, and it will not reopen. Superseded by #606 — same branch, same commit, rebased onto `main`.

PR #639, opened 2026-08-06
Compliance wave 6: stream compacting and event data masking (2.43.0)

jeremydmiller:
The two suites the 2.41.0 lifts existed to enable — marten#5153 and marten#5154. Library **22 → 24 suites, 178 → 199 tests**.

| Suite | Tests | Seam cost |
|---|---|---|
| `StreamCompactingCompliance` | 11 | **none** |
| `EventDataMaskingCompliance` | 10 | 3 members |

## `StreamCompactingCompliance` — zero seam, which is the point

`CompactStreamAsync<T>` is reached through the shared `IEventStoreOperations`, so this suite costs nothing but the file. That is exactly what the lift was for.

Compacting is supposed to be invisible from the read side, so every test pairs a **storage** assertion (how many events survive, and what the survivor is) with a **behavioural** one (what the stream still aggregates to). A store that drops the folded state, or folds it and then loses the events it should have kept, fails the second half while passing the first.

Covered: compact-at-latest, compact-at-version with the later events surviving untouched, idempotency, the no-op on a stream that does not exist, appending afterwards, the archiver callback, and the string-identified stream.

**Version preservation gets its own test** because it was the likeliest place to drift. Only Marten's own tests pinned that compacting a nine-event stream leaves the version at nine and stamps the surviving `Compacted<T>` with it. Polecat's mirror asserted only that appending afterwards still worked — which passes vacuously if the version rewound and then climbed again.

## `EventDataMaskingCompliance` — and a correction to #5154's framing

Lifting `IEventDataMasking` made the **vocabulary** shared, not the reach. Both products hand one out through `Advanced.ApplyEventDataMasking(...)` on store-specific advanced-operations types that share no interface, and register rules through their own event options. So the suite costs three seam members: `ApplyEventDataMaskingAsync` on the fixture, and the two `AddMaskingRule` overloads (`Action<T>` and `Func<T,T>`) on the config.

## ⚠️ Found a real bug — present in **both** products

`EventGraph.TryMask` ran its maskers as:

```csharp
matched = matched || masker.TryMask(e);
```

`||` short-circuits. The first rule to match an event stops every later rule from being invoked at all. Register two rules that both apply — one on an interface and one on the concrete type, say — and **only the first-registered one ever runs, while the operation still reports success**. For a right-to-erasure feature that means protected information silently survives a masking pass that looked like it worked.

The implementations are byte-identical in both repos, so this is two fixes rather than one — the code was duplicated, not shared. `rules_compose_rather_than_replace_one_another` is the test that catches it, and it failed on Marten before the fix.

## Deliberate omissions, both recorded in the suite docs

- **What an empty masking request does.** One product's own tests pin it as throwing; nothing documents that as a cross-store contract. Encoding it would pin an implementation detail rather than a promise — the same trap the `FetchLatest` suite hit over uncommitted appends.
- **The `Timestamp` cut-off on a compacting request.** Deriving one that discriminates needs two commits with server-side timestamps far enough apart to order reliably on a CI host that is not the database host, and the version cut-off already covers partial compaction deterministically.

## Verification

Marten (via `ComplianceSourceDir`): compacting **11/11**, masking **10/10**, and Marten's own 11 `removing_protected_information` tests still green alongside the product fix. Polecat verification lands with its adoption PR.

Bumps `JasperFxVersion` to **2.43.0** — a merge touching shipped code that does not bump the version is silently skipped by `--skip-duplicate`.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
     <PropertyGroup>
-        <JasperFxVersion>2.42.2</JasperFxVersion>
+        <JasperFxVersion>2.43.0</JasperFxVersion>
         <LangVersion>13</LangVersion>
         <NoWarn>1570;1571;1572;1573;1574;1587;1591;1701;1702;1711;1735;0618</NoWarn>
         <Authors>Jeremy D. Miller;Jaedyn Tonee</Authors>
--- a/src/JasperFx.Events.ComplianceTests/ComplianceStoreConfig.cs
+++ b/src/JasperFx.Events.ComplianceTests/ComplianceStoreConfig.cs
@@ -139,6 +139,30 @@ public sealed class ComplianceStoreConfig
         return this;
     }
 
+    /// <summary>
+    /// Register a mutating masking rule for an event type.
+    /// </summary>
+    /// <remarks>
+    /// Both products spell this <c>Events.AddMaskingRuleForProtectedInformation&lt;T&gt;</c> with
+    /// identical signatures, but on their own event options rather than on any shared interface,
+    /// which is the whole reason this seam member exists.
+    /// </remarks>
+    public ComplianceStoreConfig AddMaskingRule<TEvent>(Action<TEvent> rule) where TEvent : notnull
+    {
+        _registrations.Add(registrar => registrar.AddMaskingRule(rule));
+        return this;
+    }
+
+    /// <summary>
+    /// Register a replacing masking rule for an event type — the form a <c>record</c> needs.
+    /// </summary>
+    /// <inheritdoc cref="AddMaskingRule{TEvent}(Action{TEvent})" path="/remarks"/>
+    public ComplianceStoreConfig AddMaskingRule<TEvent>(Func<TEvent, TEvent> rule) where TEvent : notnull
+    {
+        _registrations.Add(registrar => registrar.AddMaskingRule(rule));
+        return this;
+    }
+
     public ComplianceStoreConfig LiveAggregation<TDoc>() where TDoc : notnull
     {
         LiveAggregations.Add(typeof(TDoc));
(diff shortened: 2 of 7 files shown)


jeremydmiller:
## Verified on both stores — 21/21 each, no capability gates

Marten and Polecat both green across the two new suites. Library is at **24 suites / 199 tests**, and both consumers now run all 199.

- Marten: full `EventSourcingTests` **1765 / 0 / 7** on net9.0
- Polecat: compliance **199/199**; full-suite run in progress

## What Polecat verification turned up

Three things, and only one of them was a test problem.

**1. A real bug — polecat#423, and a nasty one.** `StreamCompacting.ExecuteAsync` gated the archiving callback on `request.Archiver is IEventsArchiver<IDocumentOperations>`. `IEventsArchiver<T>` is **invariant**, and Polecat closes `IEventStore<,>` over `IDocumentSession`, so the archiver any JasperFx-generic caller supplies — this suite included — never matched. The callback was skipped **silently**, and compaction then went on to permanently delete the events the archiver existed to copy into cold storage first.

Marten has no equivalent because its gate and its closure agree on `IDocumentOperations`. The gate looks copied across without adjusting for the different closure; its comment still said "Polecat's callbacks close on IDocumentOperations". Fixed there to accept both closures, non-breaking.

Worth noting for the library: this path had **no test and no doc coverage in Polecat at all** before the suite arrived. `the_archiver_sees_the_events_that_are_about_to_be_replaced` is the whole of its coverage today.

**2. A gap worked around — polecat#424.** `EventStoreOptions` does not re-expose `AddMaskingRuleForProtectedInformation`, so masking rules are unreachable from the configuration surface. Its fixture goes through `StoreOptions.EventGraph` and says why. Same class as polecat#395.

**3. A test problem, now fixed here (d25f914).** `Headers["erasure"].ShouldBe("case-17")` failed on Polecat with the tell-tale `should be "case-17" but was case-17` — quoted vs unquoted, i.e. `string` vs `JsonElement`. Header values do not round-trip as the same runtime type on the two stores.

That divergence is already known and already settled: `EventMetadataCompliance` reads headers through `IEvent.GetHeader(...)` and compares `ToString()`. Followed that precedent rather than inventing a second convention — it is not what this suite is about, and it is not a defect in either store.

PR #641, opened 2026-08-07
Compliance wave 7: projection rebuild/catch-up and the dead letter path (2.44.0)

### LINKED ISSUE #5149: Compliance suite: dead letters and the projection error path
Part of #5118 (compliance wave growth), epic #5110.

`AsyncDaemonCompliance` is deliberately a smoke test: one async snapshot, one rebuild. The *error* path — a projection that throws, the event being dead-lettered, the shard continuing — is the behavior that actually costs users when it drifts, and it is declared on shared interfaces:

- `JasperFx.Events.Daemon.DeadLetterEvent`, `DeadLetterShardCount`
- `IEventDatabase.CountDeadLetterEventsAsync(ShardName)`, `QueryDeadLetterEventsAsync(ShardName, tenantId)`, `FetchDeadLetterCountsAsync()` and its per-tenant overload
- `IEventStore.AllDatabases()` → `IReadOnlyList<IEventDatabase>`
- `IEventFailureContext`

## Seam additions required

Small. `AllDatabases()` is already on the shared `IEventStore` the fixture exposes, so the dead-letter queries are reachable today. What is missing is a way for a suite to register a projection that fails deterministically *and* to configure the error policy:

1. `ComplianceStoreConfig` needs an error-handling knob — at minimum "skip poison events after N attempts and record a dead letter", which both products spell differently (Marten `OnException(...).SkipEvent()` policies on `DaemonSettings`; Polecat's equivalent). Follows the `MaxConcurrentRebuildsPerDatabase` precedent: fixture-owned property, not registrar-routed.
2. Nothing else — the failing projection itself is shared source.

## What each store tests today

| Marten | Polecat |
|---|---|
| `DaemonTests` error-handling and skip tests | `Daemon/dead_letter_count_tests.cs` |
| `marking_events_as_skipped_as_Guid_identified.cs` | `Daemon/skipped_events_count_augmenter_tests.cs` |
| `marking_events_as_skipped_as_string_identified.cs` | `Daemon/shard_failure_progression_columns_tests.cs` |
| | `Events/event_failure_context_tests.cs` |

## Scope

New `DeadLetterCompliance` suite:

1. A projection that throws on one event type dead-letters that event and the shard keeps advancing
2. `CountDeadLetterEventsAsync` / `FetchDeadLetterCountsAsync` report the shard and the count
3. `QueryDeadLetterEventsAsync` returns the event body, the exception type and the shard name
4. `IEventFailureContext` correlates the failure with the dead-letter row
5. A rebuild clears prior dead letters for the shard
6. Marking events as skipped, for Guid and string stream identity

Explicitly **out of scope**: pause/resume policies, circuit breakers, multi-node failure handling — daemon-topology behavior that stays product-specific.

## Acceptance

- One config knob added; suite enrolled in both stores
- Originals retired for the absorbed behavior in both repos

### LINKED ISSUE #5150: Compliance suite: projection rebuild and catch-up semantics
Part of #5118 (compliance wave growth), epic #5110.

`AsyncDaemonCompliance` rebuilds once, and `RebuildConcurrencyCapCompliance` pins the cap resolution. The rebuild and catch-up *semantics* are still tested independently in both repos, and they are the behaviors most often reported as bugs across both products.

Everything runs through `JasperFx.Events.Daemon.IProjectionDaemon` (`RebuildProjectionAsync`, `StartAgentAsync`, `StopAgentAsync`, `StartAllAsync`, `WaitForNonStaleData`), which the fixture already exposes via `StartDaemonAsync()` and `WaitForNonStaleProjectionDataAsync(TimeSpan)`.

## Seam additions required

Minimal — likely none for the single-tenant cases. The per-tenant rebuild cases need the tenant-scoped session and `TenancyStyle` from the conjoined tenancy issue, so sequence this one after it (or scope the per-tenant cases out and file them as a follow-up).

## What each store tests today

| Marten | Polecat |
|---|---|
| `Aggregation/rebuilding_a_single_stream_projection.cs` | `Daemon/per_tenant_rebuild_tests.cs` |
| `Aggregation/querying_with_non_stale_data.cs` | `Daemon/async_daemon_tests.cs` |
| `DaemonTests` catch-up / subscribe-from tests | `Daemon/projection_progression_tests.cs` |
| `rebuild_concurrency_cap_resolution.cs` (already lifted) | `Daemon/read_projection_progress_extended_tests.cs` |

## Scope

New `ProjectionRebuildCompliance` suite:

1. Rebuild replaces existing projected documents rather than duplicating them
2. Rebuild after the projection's logic changed produces the new shape for *old* events
3. Rebuilding one projection leaves other projections' progress untouched
4. Rebuild of an empty event store is a no-op that still completes
5. Catch-up: events appended while the daemon is stopped are processed on restart, exactly once
6. `WaitForNonStaleProjectionDataAsync` returns only after the shard has actually reached the high-water mark
7. Progress is recorded and readable after a rebuild (shared `AllProjectionProgress`)
8. Per-tenant rebuild affects only that tenant *(sequence after the conjoined tenancy issue)*

Explicitly **out of scope**: high-water detection internals, node distribution, HotCold, and the sharded/partitioned rebuild paths — all storage- or topology-specific.

## Acceptance

- Suite enrolled in both stores, zero capability gates for the single-tenant cases
- Originals retired for the absorbed behavior in both repos

jeremydmiller:
The last two event-sourcing suites in the backlog — marten#5150 and marten#5149. Library **24 → 26 suites, 199 → 216 tests**.

## Both needed zero seam additions — the backlog had them misfiled

They were grouped under "needs a seam addition." Checking the shared surfaces rather than the issue text says otherwise:

- **`IProjectionDaemon` already declares the entire rebuild surface** — seven `RebuildProjectionAsync` overloads, both `CatchUpAsync` forms, `PrepareForRebuildsAsync` — and the fixture's `StartDaemonAsync()` already hands one back.
- **The error policy and dead letters are both already shared** — `IEventStore<TOperations, TQuerySession>.ContinuousErrors` for the former, `IEventStore.AllDatabases()` → `IEventDatabase.QueryDeadLetterEventsAsync` for the latter. I verified both products *override* both rather than inheriting the empty-array DIMs before writing a line.

One wrinkle worth knowing generally: `ContinuousErrors` is on the **generic** `IEventStore<,>` while the fixture exposes the non-generic `IEventStore`. The suite casts, which is safe precisely because it is generic over the same pair the store closes over — that reaches the whole generic store surface without adding seam.

## `RebuildAndCatchUpCompliance` — 11 tests

`AsyncDaemonCompliance` already smoke-tests catch-up and a basic rebuild, so this covers what's easy to get subtly wrong: teardown-before-replay, idempotency, agreement between the by-view-type and by-name overloads, isolation from a second projection, catch-up without a full daemon run, and `PrepareForRebuildsAsync`.

**The teardown test is the one that matters.** A rebuild that replays onto surviving rows looks correct for every stream whose events still exist, and is wrong only for rows the replay can no longer produce — so it passes any test that checks a live aggregate. That exact divergence was a real product gap found by the flat-table suite (polecat#415). So this plants a document the replay cannot possibly produce and requires the rebuild to remove it.

The by-name test plants the orphan too, deliberately: a named rebuild that silently matched no projection would otherwise pass a state assertion that was already true.

## `DeadLetterCompliance` — 6 tests

A poison event whose `Apply` always throws, which needs no store-specific hook. Pins that a skip policy actually **skips** — the shard survives and keeps projecting events *after* the fault, not just before it — and that the skipped event is **recorded** rather than lost, with enough on the row to identify the event and the failure.

Corrected mid-flight: `ExceptionMessage` carries the wrapping `ApplyEventException`'s own message, not the inner reason, while `ExceptionType` names the inner exception. Asserting the inner type is what the shared ctor actually promises; searching the message for the thrown text does not work, and was my first draft.

**Deliberately not asserted:** the stop-on-error policy. How a shard pauses, stops or faults is expressed differently enough between the products that encoding one shape would pin an implementation rather than a promise.

## ⚠️ Disclosure — an intermittent failure I could not reproduce

`a_rebuild_reproduces_the_projected_state` failed **twice on Polecat** early in development, both times at ~31s, and has passed every run since — **7 consecutive full 216-test compliance runs**, including one immediately after a container restart.

It is not the rebuild shard timeout (that defaults to 5 minutes — checked). I have no error text and no reproduction, so I am not guess-fixing it or quietly hardening the test to make it go away. Recording it so it doesn't look clean: **if this resurfaces in either CI, this is the first sighting.**

## Verification

Marten 17/17 on the two new suites, full `EventSourcingTests` running. Polecat 17/17, full compliance **216/216**, full suite running.

Closes JasperFx/marten#5149
Closes JasperFx/marten#5150

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
     <PropertyGroup>
-        <JasperFxVersion>2.43.0</JasperFxVersion>
+        <JasperFxVersion>2.44.0</JasperFxVersion>
         <LangVersion>13</LangVersion>
         <NoWarn>1570;1571;1572;1573;1574;1587;1591;1701;1702;1711;1735;0618</NoWarn>
         <Authors>Jeremy D. Miller;Jaedyn Tonee</Authors>
--- a/src/JasperFx.Events.ComplianceTests/README.md
+++ b/src/JasperFx.Events.ComplianceTests/README.md
@@ -122,6 +122,8 @@ shared interfaces (`IEventStoreOperations`, `IQueryEventStore`, `IEventRegistry`
 | Strong-typed identifiers on aggregates | `StrongTypedIdentityCompliance` |
 | Stream compacting into a `Compacted<T>` snapshot | `StreamCompactingCompliance` |
 | Batch data masking of stored events | `EventDataMaskingCompliance` |
+| Projection rebuild and catch-up semantics | `RebuildAndCatchUpCompliance` |
+| The projection error path and dead letters | `DeadLetterCompliance` |
 
 ## What is deliberately out of scope
 
--- a/src/JasperFx.Events.ComplianceTests/Suites/DeadLetterCompliance.cs
+++ b/src/JasperFx.Events.ComplianceTests/Suites/DeadLetterCompliance.cs
@@ -1,0 +1,262 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using JasperFx.Events.Daemon;
+using JasperFx.Events.Projections;
+using Shouldly;
+using Xunit;
+
+namespace JasperFx.Events.ComplianceTests;
+
+#region Dead letter events and aggregate
+
+public record ReactorStarted(string Name);
+
+public record ReactorPulsed(int Output);
+
+/// <summary>
+/// Applying this event always throws. It is the whole mechanism of this suite: a poison event in an
+/// otherwise ordinary stream, so the error path can be exercised without any store-specific hook.
+/// </summary>
+public record ReactorFaulted(string Reason);
+
+public partial class ComplianceReactor
+{
+    public Guid Id { get; set; }
+    public string Name { get; set; } = string.Empty;
+    public int Output { get; set; }
+
+    public static ComplianceReactor Create(ReactorStarted e) => new() { Name = e.Name };
+
+    public void Apply(ReactorPulsed e) => Output += e.Output;
+
+    public void Apply(ReactorFaulted e)
+        => throw new InvalidOperationException($"Reactor fault: {e.Reason}");
+}
+
+#endregion
+
+/// <summary>
+/// The projection error path — what happens to an async projection when applying an event throws,
+/// and what the store records about it.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Two behaviours are worth pinning across stores and neither is obvious. First, that a skip policy
+/// actually <em>skips</em>: the shard survives the poison event and keeps projecting the events
+/// after it, rather than stopping or silently dropping the rest of the stream. Second, that the
+/// skipped event is <em>recorded</em> rather than lost — a dead letter row a human or a monitoring
+/// tool can find, carrying enough to identify the event and the failure.
+/// </para>
+/// <para>
+/// The error policy is reached through <c>IEventStore&lt;TOperations, TQuerySession&gt;.ContinuousErrors</c>,
+/// and the dead letters through <c>IEventStore.AllDatabases()</c> →
+/// <c>IEventDatabase.QueryDeadLetterEventsAsync</c>. Both are already shared, so this suite needs no
+/// seam addition — the reason it moved out of the "needs a seam" group (marten#5149). The cast from
+/// the fixture's non-generic <see cref="IEventStore"/> to the closed generic is safe because this
+/// suite is generic over the same pair the store closes over.
+/// </para>
+/// <para>
+/// Deliberately not asserted: the stop-on-error policy. Whether a shard pauses, stops, or faults —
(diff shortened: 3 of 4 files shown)


jeremydmiller:
## Refinement on the disclosed intermittent failure

Both sightings were **Polecat only**. Since writing the PR body:

- **Marten**: 5 consecutive full compliance runs, **216/216** each, ~14s per run. Zero sightings.
- **Polecat**: 7 consecutive full compliance runs, **216/216** each, ~100s+ per run. Zero sightings since the two early ones.

The asymmetry is itself informative. The suite logic is identical on both stores, so a fault confined to one of them points at either a Polecat-side race under concurrency or SQL Server timing, rather than anything wrong with the test — which is the main reason I did not want to "fix" it blind by hardening the assertion.

Still no error text and still no reproduction, so this remains a first sighting to watch rather than a diagnosis.

Issue #462, opened 2026-06-18
AggregateEvolverGenerator double-emits when bundled in two referenced packages (Marten + Polecat) → CS0111


jeremydmiller:
## Summary

`JasperFx.Events.SourceGenerator` is bundled (as a built-in analyzer) inside **both** the Marten and the Polecat NuGet packages (`analyzers/dotnet/cs/JasperFx.Events.SourceGenerator.dll`). A project that references **both** stores loads the *same* incremental generator **twice**. `AggregateEvolverGenerator` emits each projection's evolver as partial members of the user's class; with two analyzer instances, those members are emitted twice and the consumer's compile fails with:

```
error CS0111: Type 'AlwaysFailsProjection' already defines a member called 'AlwaysFailsProjection' ...
error CS0111: Type 'AlwaysFailsProjection' already defines a member called 'Evolve' ...
```

## Why it can't be fixed within the generator

`AggregateEvolverGenerator.ExecuteCombined` dedupes thoroughly **within a single analyzer instance** (`directEmitted` / `seen` / `pipeline3Pairs` sets, unique hint-name guards). But two analyzer **instances** cannot coordinate — each incremental generator only ever sees the *original* compilation, never the other instance's added sources. So no amount of in-generator dedup logic helps; both instances independently decide to emit `Foo.Evolver.g.cs` with the same members.

## Why it surfaced now

Previously the two bundled copies were *different* generator versions (e.g. Marten bundled 2.13.0, Polecat bundled 2.12.0), and the differing output/identities didn't collide on these projection types. Once **Marten 9.9.1 and Polecat 4.5.1 both shipped the identical 2.13.1 generator**, the two instances emit byte-identical partials → CS0111. The generator code itself is unchanged between 2.13.0 and 2.13.1 (`git diff V2.13.0 V2.13.1 -- src/JasperFx.Events.SourceGenerator/` is empty) — this is purely a double-load/packaging interaction, not a code regression.

## Repro

A single project (e.g. a test project) that references both `Marten` and `Polecat` and declares a `SingleStreamProjection<TDoc, TId>` subclass. Both bundled analyzers run → duplicate evolver members → CS0111.

## Current consumer-side workaround (CritterWatch)

Drop one bundled copy in the consuming project, keeping the other (the generator is store-agnostic, so one instance still emits dispatchers for every projection in the compilation):

```xml
<Target Name="DropDuplicateBundledEventSourceGenerator" BeforeTargets="CoreCompile">
[...]

PR #651, opened 2026-08-13
fix(source-generator): only require `partial` where the dispatcher lives in the user's class

### LINKED ISSUE #650: Projection subclasses are silently skipped unless `partial`, though only the injected-dispatcher paths need it
## Symptom

A projection subclass with conventional methods and no `partial`:

```csharp
public class FooProjection : SingleStreamProjection<FooDoc, FooId>
{
    public static FooDoc Create(FooCreated e) => new() { ... };
    public static void Apply(IFooChanged e, FooDoc doc) { ... }
}
```

- `dotnet build` → 0 errors, 0 warnings.
- No evolver is emitted — zero generated files for that type; a `partial` control in the same assembly emits one.
- At `DocumentStore` construction: `InvalidProjectionException: No source-generated dispatcher found for FooProjection … there is no runtime fallback`.

In a modular codebase every real-host integration test in every module dies at fixture init, with a stack trace pointing at store construction rather than at the projection. It also gets reintroduced silently: ReSharper/Rider's redundant-partial inspection removes the modifier — correctly, by the language's own rules — so a projection written `partial` becomes non-partial on the next cleanup-on-save.

Reproduced on Marten 9.17.1 and 9.23.0.

## Cause

`AggregateAnalyzer.AnalyzeProjectionSubclass` gates candidacy on the modifier (`src/JasperFx.Events.SourceGenerator/AggregateAnalyzer.cs`, ~line 241):

```csharp
var isPartial = classDecl.Modifiers.Any(SyntaxKind.PartialKeyword);
...
Mode = isPartial ? CandidateMode.PartialProjection : CandidateMode.None,
```

The emission that gate was protecting no longer exists on this path. Per the doc comment on `EvolverCodeEmitter.EmitPartialProjection`, #462 moved aggregation dispatch from members injected into the user's class to a standalone `file sealed class` registered via `[assembly: GeneratedEvolver(...)]`, precisely so a twice-loaded generator (Marten + Polecat) cannot trip CS0111. The emitted evolver reaches the projection through public static calls — or, for instance methods, through its own `new` — and never needs a second part. The candidacy gate was not narrowed when the emission changed; `git log` confirms it is unchanged from the original generator commit.

`IsCandidateClass` already argues exactly this for records ("without ever requiring the record to be `partial`", marten#4557).

## Where `partial` is genuinely required

Not everything can move. These paths still emit into the user's own class and must keep requiring the modifier:

1. **EventProjection subclasses** (`AnalyzeEventProjectionSubclass`, ~line 978, and the explicit-`ApplyAsync` path at ~line 944). The dispatcher *is* an `ApplyAsync` override on the user's class — `EmitEventProjectionPartial`.
2. **marten#4787** — instance conventional methods with no public parameterless constructor. The file-scoped evolver's shadow instance is built with `RuntimeHelpers.GetUninitializedObject`, so injected fields are null. Also covers an abstract projection base class, whose implicit constructor is protected.
3. **Generic or non-visible projections** — see below.

## Secondary: JFXEVT003 is wrong in both halves

`DiagnosticDescriptors.NotPartial`:

```csharp
messageFormat: "Projection '{0}' is not partial; falling back to runtime expression compilation",
defaultSeverity: DiagnosticSeverity.Info,
```

- **The message is false.** There is no runtime fallback. `AggregateApplication` is validity-only; `AggregateApplication.cs:123` and `JasperFxAggregationProjectionBase.Runtime.cs:153` both throw *"there is no runtime fallback"*. Marten 9.23.0's own `Marten.csproj` says the same, and `Bug_4557_self_aggregating_snapshot_without_source_generator` asserts the exception text.
- **Info is invisible.** Verified against a real consumer build: the diagnostic does not appear at `-v n` *or* at `-v d`. At best it is an IDE hint — and the IDE is what strips the modifier in the first place.

It also fires where nothing is wrong: for a projection whose conventional methods live on the aggregate (that projection dispatches fine through the aggregate's own evolver), and for a *partial* projection using constructor lambda registrations (`IsPartial` is never set on that `CandidateInfo`).

## Also found while tracing this

Both reproduce on `main` **with** `partial`, so they are independent of the gate — but they are on the same code path and a fix has to account for them:

- **A generic projection subclass emits uncompilable code.** `typeof(global::Ns.FooProjection<T>)` inside a non-generic file-scoped evolver → CS0246.
- **A projection nested behind `private`/`protected` emits uncompilable code.** `file`-scoped types cannot be nested, so the evolver cannot name it → CS0122.

Both compile correctly when routed to the member-injection emission instead, which handles generics and containing types — provided the projection and every containing type are `partial` (a non-partial container is CS0260).

## Proposed fix

1. `AnalyzeProjectionSubclass` sets `CandidateMode.PartialProjection` regardless of the modifier, keeping `IsPartial` on the `CandidateInfo`.
2. Before emitting, check whether this projection needs the dispatcher inside its own class (marten#4787 shape, or not nameable from a separate file). If so and a `partial` modifier is missing — on the projection or any containing type — report JFXEVT003 naming that declaration and emit nothing.
3. Route generic / non-visible projections to the member-injection emission so they work at all.
4. Rescope JFXEVT003 to those cases, raise it to **Error**, and give it an accurate message. The consequence is a hard runtime failure, not a performance degradation.

## Validation notes for a fix

- `AggregateEvolverGeneratorTests.skips_non_partial_projection` encodes the current behaviour and has to change.
- The #462 double-load scenario must stay clean. Simulating it needs two *distinct* generator types over one compilation — two instances of the same type collide on generated file path alone (CS0433), which is not what #462 was about.
- The marten#4787 path must still emit its override and still refuse non-partial loudly.

Blackclaws:
## What

Narrows the `partial` requirement on projections to the paths that actually need it, and makes the diagnostic for those paths tell the truth.

Closes #650

## Why

`AnalyzeProjectionSubclass` mapped every non-partial aggregation projection subclass to `CandidateMode.None`, so nothing was emitted: build clean, then `InvalidProjectionException: No source-generated dispatcher found` at `DocumentStore` construction. The gate is unchanged from the original generator commit, when dispatch *was* members injected into the user's class. #462 moved aggregation dispatch to a standalone `file sealed class` registered through `[assembly: GeneratedEvolver(...)]` — which calls the projection's public methods from the outside and needs nothing from the user's declaration — but the gate was never narrowed to match. `IsCandidateClass` already makes this argument for records (marten#4557).

## Changes

**Candidacy no longer looks at the modifier.** `AnalyzeProjectionSubclass` always yields `CandidateMode.PartialProjection`; `IsPartial` is still carried for the emitter.

**`partial` is now required only where the dispatcher is written into the projection class itself**, reported as `JFXEVT003` naming the declaration that lacks it:

| shape | why it needs `partial` |
|---|---|
| `EventProjection` subclass with conventional methods | its dispatcher *is* an `ApplyAsync` override on the user's class |
| instance methods + no public parameterless ctor | dispatch has to run on the DI-built instance (marten#4787); the evolver's `GetUninitializedObject` shadow leaves injected fields null. Also covers abstract projection base classes |
| generic projection, or one behind `private`/`protected` | a separate `file`-scoped type cannot name it |

**Generic and non-visible projections now work.** They previously emitted `typeof(Foo<T>)` (CS0246) or an unreachable type reference (CS0122) *even when declared `partial`* — a build break in generated code the consumer cannot edit. They now route to the member-injection emission, which already handles type parameters and containing types. A non-partial containing type would produce CS0260, so that is checked and reported as `JFXEVT003` too.

**JFXEVT003 rescoped, corrected, and raised to Error.** It said *"falling back to runtime expression compilation"* at `Info`. No such fallback exists — `AggregateApplication` is validity-only and both `AggregateApplication.cs` and `JasperFxAggregationProjectionBase.Runtime.cs` throw *"there is no runtime fallback"* — and `Info` diagnostics never appear in a CLI build, verified at both `-v n` and `-v d`. It also no longer fires where nothing is wrong: for a projection whose conventional methods live on the aggregate, or for a *partial* projection using constructor lambda registrations.

**Runtime messages and docs** updated to the narrower rule (`AggregateApplication.MissingDispatcherMessage`, the `EvolveAsync` backstop, `docs/migration-guide.md`, and the stale mode table in `docs/codegen/projection-apply-dispatch.md`, which still described the pre-#462 emission).

## Test plan

- `JasperFx.Events.SourceGenerator.Tests`: 46 passed. `skips_non_partial_projection` encoded the old behaviour and is removed; `PartialRequirementTests` covers its replacement plus the static-method shape, the DI shape (error), generic and private-nested projections (now emit and compile), a non-partial container (error naming it), the EventProjection case (still an error), and a double-load regression test.
- `OwnDispatchOverrideTests` pins the paths that must NOT reach the new error: a projection supplying its own `Evolve` / `EvolveAsync` override is skipped before candidacy, including the DI-activated shape that would otherwise take the JFXEVT003 branch.
- New tests live in their own files rather than at the end of `AggregateEvolverGeneratorTests`, sharing a `GeneratorHarness` whose content is identical across the related PRs so they do not collide on one test file at merge time.
- `EventTests`: 746 passed on net9.0 and net10.0. `EventStoreTests`: 114 passed on both.
- **#462 double-load re-verified.** Two *distinct* generator types over one compilation — two instances of the same type collide on generated file path alone (CS0433), which is unrelated to double loading. Every file-scoped emission is clean; the pre-existing member-injection path (marten#4787) still trips CS0111 under a double load, unchanged by this PR and filed separately.
- **Marten 9.23.0 corpus.** Built `src/EventSourcingTests` with this generator swapped in for the published analyzer: 0 errors, no `JFXEVT` diagnostics, and the generated output is byte-identical to the published generator's across all 245 files. Every projection in Marten's suite is already `partial`, so this is a no-regression result rather than a behaviour change there.

## Breaking change

`JFXEVT003` goes from `Info` (invisible) to `Error`. A projection that needs `partial` and does not have it now fails the build instead of failing at store construction. Every affected consumer is already broken at runtime today; the change moves the failure earlier and names the fix. Suppressible per-project with `dotnet_diagnostic.JFXEVT003.severity` if anyone needs a staged upgrade.

## Pipeline dedupe

`MarkSeen` adds the aggregate type to the cross-pipeline `seen` set for any `PartialProjection` candidate, which now includes non-partial subclasses, so it is worth stating why that changes nothing. When the aggregate is declared in the same compilation, pipeline 1 emits its self-aggregating evolver from its own declaration and both registrations are present — verified identical with and without this change. When it is declared in another assembly, pipelines 2 and 3 never emitted for it in the first place: `AnalyzeRefersToAggregate` requires a `DeclaringSyntaxReference` (AggregateAnalyzer.cs:1633) and returns null without one. Verified end to end by compiling the aggregate into a referenced assembly and registering `Snapshot<T>` in the host — nothing is emitted there either way, which is exactly why marten#4557 was fixed by running the generator in the aggregate's own assembly.

### DIFF
--- a/docs/codegen/projection-apply-dispatch.md
+++ b/docs/codegen/projection-apply-dispatch.md
@@ -31,7 +31,7 @@ The SG emits one of five shapes through `EvolverCodeEmitter`
 
 | Mode | Emits | How runtime picks it up |
 |---|---|---|
-| `PartialProjection` | Partial method on the user's projection class (`Evolve` / `EvolveAsync` / `DetermineActionAsync`) | `JasperFxAggregationProjectionBase.isOverridden(...)` returns `true` for the generated override and `_usesConventionalApplication = false` — `AggregateApplication` is never invoked for dispatch. |
+| `PartialProjection` | Standalone `file`-scoped evolver + assembly-level `[GeneratedEvolverAttribute]` carrying the projection type. Falls back to an `Evolve` / `EvolveAsync` / `DetermineActionAsync` override injected into the user's class when the evolver cannot do the job — dispatch has to run on a DI-built instance (marten#4787), or the projection is generic / not visible from another file. | `tryUseAssemblyRegisteredEvolver(...)` binds the registered evolver; for the injected-override fallback, `isOverridden(...)` returns `true` first and `_usesConventionalApplication = false`. Either way `AggregateApplication` is never invoked for dispatch. |
 | `SelfAggregating` | Standalone evolver class implementing `IGeneratedSyncEvolver<TDoc,TId>` or `IGeneratedSyncDetermineAction<TDoc,TId>` + an assembly-level `[GeneratedEvolverAttribute]` | `JasperFxAggregationProjectionBase.tryUseAssemblyRegisteredEvolver(...)` activates and binds the evolver. |
 | `SelfAggregatingEvolve` | Standalone evolver class implementing `IGeneratedAsyncEvolver<TDoc,TId>` + `[GeneratedEvolverAttribute]` | Same as above. |
 | `EventProjection` | Partial `ApplyAsync(TOperations, IEvent, CancellationToken)` override on the user's `JasperFxEventProjectionBase<TOperations>` subclass | The override wins at vtable dispatch. `EventProjectionApplication` is only used when the base `ApplyAsync` is not overridden. |
@@ -156,6 +156,14 @@ emit shape or a reflective rewrite.
    (e.g. those that override `ApplyAsync` or `Evolve*` directly) do not need
    to be `partial`. The fail-fast exception message at registration spells
    out this narrower condition.
+
+   > Narrowed again since. Moving the aggregation dispatcher to a file-scoped
+   > evolver (#462) removed the requirement for `SingleStreamProjection` /
+   > `MultiStreamProjection` subclasses as well — it survives only where the
+   > dispatcher still has to be written into the user's class: an
+   > `EventProjection`, a projection dispatching on a DI-built instance
+   > (marten#4787), and generic or non-visible projections. Reported as
+   > `JFXEVT003` at build time.
 
 ## What shipped in #276
 
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -79,27 +79,35 @@ public Task ApplyAsync(TOperations operations,
 
 > `ProjectionLifecycle` (the `Inline` / `Async` / `Live` enum passed to `Projections.Add<T>(...)`) moved to the same `JasperFx.Events.Projections` namespace for the same reason. `using JasperFx.Events.Projections;` covers both.
 
-#### Convention-based projection subclasses must be declared `partial`
+#### Convention-based projections are dispatched by the source generator
 
-Conventional `Apply` / `Create` / `ShouldDelete` methods on a **projection subclass** — one that derives from `SingleStreamProjection<TDoc, TId>`, `MultiStreamProjection<TDoc, TId>`, or `EventProjection` — are now dispatched by the compile-time `JasperFx.Events.SourceGenerator`. Pre-2.0 they were dispatched by runtime reflection; **in 2.0 there is no runtime reflection fallback.** For the generator to emit its `[GeneratedEvolver]` dispatcher, the projection class must be declared `partial`, and its convention methods must be `public`.
+Conventional `Apply` / `Create` / `ShouldDelete` methods on a **projection subclass** — one that derives from `SingleStreamProjection<TDoc, TId>`, `MultiStreamProjection<TDoc, TId>`, or `EventProjection` — are now dispatched by the compile-time `JasperFx.Events.SourceGenerator`. Pre-2.0 they were dispatched by runtime reflection; **in 2.0 there is no runtime reflection fallback.** The convention methods must be `public` so the generated dispatcher can call them.
 
 ```csharp
-// before (JasperFx.Events 1.x) — reflection dispatched the conventional methods
 public class TripProjection : SingleStreamProjection<Trip, Guid>
-{
-    public Trip Create(IEvent<TripStarted> e) => new() { Id = e.StreamId };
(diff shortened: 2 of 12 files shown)


Blackclaws:

Issue #297, opened 2026-05-18
AggregateEvolverGenerator alpha.4: handler-argument binding, duplicate switch cases, and required-member init bugs


jeremydmiller:
## Summary

Following the jasperfx#292 fix that landed in `JasperFx.Events.SourceGenerator` 2.0.0-alpha.4, the SG now correctly handles nested projection classes and self-aggregating records — but the generated `Evolver.g.cs` files have three new categories of codegen bugs that surface as build failures in the Marten test suite. 128 compile errors across ~14 unique projection types in `src/EventSourcingTests`.

## Category 1 — CS1503 handler-argument binding (48 errors)

The SG emits an invocation of an `Apply` / `Create` handler where the positional arguments are bound to the wrong types. Looks like the SG is picking the same value (the event, or the aggregate) and passing it for two different parameters.

**Repro A** (extra `IDocumentOperations` parameter; SG passes the aggregate where ops are expected):

```csharp
public partial class InvalidArgumentType : SingleStreamProjection<MyAggregate, Guid>
{
    public void Apply(AEvent @event, MyAggregate aggregate, IDocumentOperations operations) { /* ... */ }
}
```

Generates:
```
error CS1503: Argument 3: cannot convert from 'MyAggregate' to 'Marten.IDocumentOperations'
```

**Repro B** (`IQuerySession` parameter; SG passes the event where session is expected):

```csharp
public partial class CanGuessEventType : SingleStreamProjection<MyAggregate, Guid>
{
    public void Apply(AEvent a, MyAggregate aggregate, IQuerySession session) { /* ... */ }
}
```

Generates:
```
error CS1503: Argument 3: cannot convert from 'AEvent' to 'Marten.IQuerySession'
```

**Repro C** (`Apply` on a self-aggregating record; SG mis-binds the event and the current-state args):

```csharp
public record CounterWithCreateAndDefaultCtor(Guid Id, int CreateCounter, int ApplyCounter)
{
    public CounterWithCreateAndDefaultCtor() : this(Guid.Empty, 0, 0) { }

    public static CounterWithCreateAndDefaultCtor Create(IncrementEvent @event, IEvent metadata) => new(metadata.StreamId, 1, 0);

    public CounterWithCreateAndDefaultCtor Apply(UnrelatedEvent e, CounterWithCreateAndDefaultCtor current) => current;
    public CounterWithCreateAndDefaultCtor Apply(IncrementEvent @event, CounterWithCreateAndDefaultCtor current)
[...]

Issue #286, opened 2026-05-18
Remove inline-lambda projection registration APIs in 2.0


jeremydmiller:
Follow-up from #276. The inline-lambda registration APIs on the projection /
aggregator base classes still rely on FastExpressionCompiler-compiled
delegates, because the source generator cannot statically discover handlers
that are passed as runtime values:

- `AggregateApplication.CreateEvent<TEvent>(handler)`
- `AggregateApplication.DeleteEvent<TEvent>(handler)`
- `AggregateApplication.ProjectEvent<TEvent>(handler)` (7 overloads)
- `EventProjectionApplication.Project<TEvent>(action)`
- `EventProjectionApplication.ProjectAsync<TEvent>(action)`

Goal: remove these APIs entirely in 2.0. Consumers express the same intent by
implementing `Apply` / `Project` / `ApplyAsync` as methods on a `partial`
projection class, which the source generator dispatches to.

Scope:
- Delete the registration methods and their internal FEC dispatch sites
  (`AggregateApplication.Register.cs` x 3, `EventProjectionApplication.projectEvent`).
- Drop the `IL2026` / `IL3050` class-level suppressions that justified them.
- Drop the `FastExpressionCompiler` package reference if nothing else in
  `JasperFx.Events` still needs it (the strong-typed-id converters in
  `IEvent.cs` and `StreamAction.cs` still use FEC and remain on it for now).
- Document the migration in the 2.0 release notes.

Coordinated with the matching Marten 9.0 migration note (see JasperFx/marten companion issue).

PR #657, opened 2026-08-13
refactor(source-generator): drop the opt-out for removed lambda registration APIs

### LINKED ISSUE #655: JFXEVT001/JFXEVT004 are never reported, and the lambda opt-out they describe now only produces false skips
Two of the five `DiagnosticDescriptors` are never passed to `ReportDiagnostic` anywhere in the generator:

- `JFXEVT001` `AsyncSelfAggregating` — "Self-aggregating type '{0}' has async methods; falling back to runtime expression compilation"
- `JFXEVT004` `HasLambdaRegistrations` — "Skipping '{0}' — has lambda registrations in constructor"

`JFXEVT001`'s message is also stale in the same way JFXEVT003's was (see #650): there is no runtime expression-compilation fallback, and async self-aggregating handlers have been supported via `IGeneratedAsyncEvolver` since #297, so the condition it describes is no longer a skip at all.

## The JFXEVT004 case is worse than dead

The detection it belongs to is still live. `AggregateAnalyzer.HasLambdaRegistrations` scans a projection's constructor for `ProjectEvent` / `CreateEvent` / `DeleteEvent` invocations **with arguments** and, on a match, returns `CandidateMode.None` — no dispatcher, and no diagnostic since JFXEVT004 is never reported.

Those inline-lambda APIs no longer exist. They were removed in JasperFx 2.0 / Marten 9.0 (#286); Marten's docs direct users to method conventions instead:

- `docs/events/projections/conventions.md` — "The pre-9.0 `ProjectEvent<T>(...)` / `ProjectEventAsync<T>(...)` constructor helpers were removed"
- `docs/events/projections/event-projections.md` — *Removed in Marten 9.0* warning for `Project<TEvent>(action)` / `ProjectAsync<TEvent>(action)`
- `docs/migration-guide.md` — the removal table, all 7 `ProjectEvent` overloads

What survives on the aggregation base is `DeleteEvent<TEvent>()` and `TransformsEvent<TEvent>()`, both parameterless, and the check requires at least one argument — so it can no longer match any supported API.

The only remaining way to trigger it is a **user-defined helper** named `ProjectEvent` / `CreateEvent` / `DeleteEvent` called with arguments from a projection constructor. That silently suppresses the dispatcher and surfaces later as `InvalidProjectionException: No source-generated dispatcher found` at store construction.

## Suggested resolution

Drop `HasLambdaRegistrations` (analysis and descriptor) along with `JFXEVT001`, or — if the opt-out is still wanted for the removed APIs during migration — bind it to the actual framework symbols rather than method names, and report it so the skip is visible.

Blackclaws:
Closes #655

## Why

`AggregateAnalyzer.HasLambdaRegistrations` skipped any projection whose constructor invoked `ProjectEvent` / `CreateEvent` / `DeleteEvent` with arguments, and its EventProjection counterpart matched the raw substring `"Project<"` over the constructor's text.

Those inline-lambda registration APIs were removed in JasperFx 2.0 / Marten 9.0 (#286). Marten's docs say so in three places and route users to method conventions instead. What survives on the aggregation base is `DeleteEvent<TEvent>()` and `TransformsEvent<TEvent>()`, both parameterless — and the check requires at least one argument, so it can no longer match a supported API at all.

That leaves only false positives: a user-defined helper that happens to share one of those names, or — for EventProjection — a mere mention of `Project<` in a comment. When it fired, the projection got no dispatcher and nothing was reported, surfacing later as `No source-generated dispatcher found` at store construction.

## Changes

Both detectors removed, along with `JFXEVT004` (which described this skip) and `JFXEVT001` — neither was ever passed to `ReportDiagnostic`, and `JFXEVT001` also promised the runtime expression-compilation fallback that no longer exists, for a condition (async self-aggregating handlers) that has been supported since #297.

## Test plan

`LambdaOptOutRemovalTests` pins both false positives: a projection with its own `ProjectEvent(string)` helper, and an EventProjection whose constructor only mentions `Project<` in a comment. Both now get dispatchers.

Source generator tests 36 passed; EventTests 746 on net9.0 and net10.0; EventStoreTests 114.

### DIFF
--- a/src/JasperFx.Events.SourceGenerator.Tests/GeneratorHarness.cs
+++ b/src/JasperFx.Events.SourceGenerator.Tests/GeneratorHarness.cs
@@ -1,0 +1,110 @@
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace JasperFx.Events.SourceGenerator.Tests;
+
+/// <summary>
+/// Shared compilation harness for the generator's regression suites.
+///
+/// <para>Each issue's tests live in their own file rather than all landing at the end of
+/// AggregateEvolverGeneratorTests, so independent fixes to the generator do not collide on one test
+/// file. This type is what makes that practical — the content is deliberately identical wherever it
+/// appears, so branches that both introduce it merge without a conflict.</para>
+/// </summary>
+internal static class GeneratorHarness
+{
+    private static List<MetadataReference> References()
+    {
+        var references = new List<MetadataReference>
+        {
+            MetadataReference.CreateFromFile(typeof(object).Assembly.Location),
+            MetadataReference.CreateFromFile(typeof(IEvent).Assembly.Location),
+            MetadataReference.CreateFromFile(typeof(Guid).Assembly.Location),
+        };
+
+        var runtimeDir = System.IO.Path.GetDirectoryName(typeof(object).Assembly.Location)!;
+        references.Add(MetadataReference.CreateFromFile(System.IO.Path.Combine(runtimeDir, "System.Runtime.dll")));
+        references.Add(MetadataReference.CreateFromFile(System.IO.Path.Combine(runtimeDir, "System.Collections.dll")));
+
+        return references;
+    }
+
+    private static CSharpCompilation Compilation(string source)
+    {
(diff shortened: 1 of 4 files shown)


Blackclaws:

PR #658, opened 2026-08-13
fix(source-generator): report unregistered published types instead of skipping in silence

### LINKED ISSUE #654: Non-partial EventProjection with an explicit ApplyAsync loses published-type registration with no diagnostic
`AggregateAnalyzer.AnalyzeEventProjectionSubclass` bails on a non-partial EventProjection that has an explicit `ApplyAsync` override (`AggregateAnalyzer.cs`, ~line 944):

```csharp
if (HasApplyAsyncOverride(classSymbol))
{
    // ... discover Store/Insert/Delete document types for published-type registration (marten#4166)
    if (!isPartial) return null;
```

Returning `null` means no candidate at all, so unlike every other skip in the generator this one produces **no diagnostic whatsoever** — not even the Info-level JFXEVT003. The published types discovered from `Store<T>` / `Insert<T>` / `Delete<T>` calls are silently not registered.

## Reproduction

```csharp
public class NonPartialExplicitApply : TestEventProjection   // note: no `partial`
{
    public override ValueTask ApplyAsync(ITestOperations operations, IEvent e, CancellationToken cancellation)
    {
        operations.Store(new AuditRecord());
        return default;
    }
}
```

Generated files: 0. Diagnostics: none.

## Impact

Low but entirely silent. Per the generator's own comments the consequences are limited to ahead-of-time surfaces — schema creation, known document types, rebuild teardown — because the store still provisions the document's storage on demand.

## Note on fixing

Emission is not the fix here: the registration is a `PublishedTypes()` override written into the user's class, so it genuinely requires `partial`. A diagnostic naming the projection is the whole remedy.

Blackclaws:
Closes #654

## Why

`AnalyzeEventProjectionSubclass` bailed on a non-partial EventProjection with an explicit `ApplyAsync` override:

```csharp
if (HasApplyAsyncOverride(classSymbol))
{
    if (!isPartial) return null;   // no candidate, no diagnostic, no registration
```

Returning `null` means no candidate at all, so unlike every other skip in the generator this one produced **no diagnostic whatsoever**. The document types discovered from the projection's `Store` / `Insert` / `Delete` calls were silently never registered as published types.

## Changes

Discovery now runs regardless of the modifier, and the emission step reports **JFXEVT006** naming the projection and the document types that went unregistered.

Emission still requires `partial` — the registration is a `PublishedTypes()` override written into the user's class, so a non-partial projection genuinely cannot receive one. The fix is the diagnostic, not new emission. The per-operation `JFXEVT005` notes are suppressed in that case: nothing is being registered for the projection at all, so pointing at individual calls would only add noise.

**Severity is Warning, not Error.** Nothing fails at runtime — the store provisions the document's storage on demand, and only the ahead-of-time surfaces (schema creation, known document types, rebuild teardown) come up short. That is also why this went unnoticed.

## Test plan

`PublishedTypeRegistrationDiagnosticTests`: reports with the document type named; stays silent when there is nothing registrable; the partial path still registers as before.

Source generator tests 37 passed; EventTests 746 on net9.0 and net10.0; EventStoreTests 114.

### DIFF
--- a/src/JasperFx.Events.SourceGenerator.Tests/GeneratorHarness.cs
+++ b/src/JasperFx.Events.SourceGenerator.Tests/GeneratorHarness.cs
@@ -1,0 +1,110 @@
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace JasperFx.Events.SourceGenerator.Tests;
+
+/// <summary>
+/// Shared compilation harness for the generator's regression suites.
+///
+/// <para>Each issue's tests live in their own file rather than all landing at the end of
+/// AggregateEvolverGeneratorTests, so independent fixes to the generator do not collide on one test
+/// file. This type is what makes that practical — the content is deliberately identical wherever it
+/// appears, so branches that both introduce it merge without a conflict.</para>
+/// </summary>
+internal static class GeneratorHarness
+{
+    private static List<MetadataReference> References()
+    {
+        var references = new List<MetadataReference>
+        {
+            MetadataReference.CreateFromFile(typeof(object).Assembly.Location),
+            MetadataReference.CreateFromFile(typeof(IEvent).Assembly.Location),
+            MetadataReference.CreateFromFile(typeof(Guid).Assembly.Location),
+        };
+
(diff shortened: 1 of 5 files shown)


Blackclaws:

PR #659, opened 2026-08-13
fix(events): only treat a genuine override of the dispatch virtuals as an override

### LINKED ISSUE #656: A projection method that hides (rather than overrides) Evolve is treated as an override, losing dispatch
`isOverridden` treats a method that merely *hides* the base virtual as an override, so a projection that declares `Evolve` without the `override` keyword takes the user-override path and dispatch is lost.

`JasperFxAggregationProjectionBase.isOverridden` (`JasperFxAggregationProjectionBase.cs:160`) asks only where the method was declared:

```csharp
private bool isOverridden(string methodName)
{
    return GetType().GetMethod(methodName)!.DeclaringType!.Assembly != typeof(IEvent).Assembly;
}
```

A hiding member — `new TDoc? Evolve(...)`, or the same signature written without any modifier, which the compiler already flags as CS0114 — is declared on the user's class, so this returns `true`. But hiding is not overriding: the base virtual still resolves to the base implementation, and the generated dispatcher is skipped because `establishBuildActionAndEvolve` checks `isOverridden` before `tryUseAssemblyRegisteredEvolver`.

## Two failure modes, both verified

**With conventional methods** — registration fails, describing a conflict the user does not have:

```csharp
public partial class HidesEvolveWithConventions : SingleStreamProjection<MyAggregate, Guid>
{
    public new MyAggregate? Evolve(MyAggregate? snapshot, Guid id, IEvent e) { ... }

    public void Apply(AEvent e, MyAggregate a) => a.ACount++;
}
```

```
JasperFx.Events.Projections.InvalidProjectionException:
This projection can only use the override of 'Evolve' or conventional
Apply/Create/ShouldDelete methods, but not both
```

Nothing was overridden, and the `Apply` method has a perfectly good generated evolver that is never consulted.

**Without conventional methods** — registration succeeds and the failure lands at first event dispatch instead:

```
JasperFx.Events.Daemon.ApplyEventException: Failure to apply event #1 Id(...)
 ---> System.NotImplementedException: Did you forget to implement this?
```

That inner message comes from the base `Evolve` at `JasperFxAggregationProjectionBase.Runtime.cs:177` — the virtual call never reaches the user's hiding method.

## Suggested fix

Require a genuine override. `MethodInfo.GetBaseDefinition()` returns the base-class declaration for an override and the method itself for a hiding member:

```csharp
private bool isOverridden(string methodName)
{
    var method = GetType().GetMethod(methodName)!;
    return method.DeclaringType!.Assembly != typeof(IEvent).Assembly
           && method.GetBaseDefinition().DeclaringType!.Assembly == typeof(IEvent).Assembly;
}
```

With that, a hiding `Evolve` stops claiming the dispatch path: the source-generated evolver binds as it should, and a projection with no dispatcher at all gets the accurate "No source-generated dispatcher found" fail-fast at registration rather than `NotImplementedException` at first event. The user still gets CS0114 from the compiler telling them to add `override` if that is what they meant.

## Adjacent, same method

`GetType().GetMethod(name)` throws `AmbiguousMatchException` when the projection declares more than one method with that name — e.g. a private `Evolve(string)` helper alongside the real one. Worth a `GetMethods()` filter in the same change.

Blackclaws:
Closes #656

## Why

`isOverridden` asked only where a method was declared:

```csharp
return GetType().GetMethod(methodName)!.DeclaringType!.Assembly != typeof(IEvent).Assembly;
```

A method that merely *hides* the base virtual — declared without `override`, which the compiler already reports as CS0114 — is declared on the user's class, so this returned `true`. Hiding is not overriding: the virtual call still lands on the base implementation, and because this check runs before `tryUseAssemblyRegisteredEvolver`, the generated dispatcher was skipped too.

Both outcomes were silent about the real cause, and both are verified in the issue:

- **with conventional methods alongside** — registration threw *"This projection can only use the override of 'Evolve' or conventional Apply/Create/ShouldDelete methods, but not both"*, a conflict the author never wrote, while their `Apply` had a perfectly good evolver going unused;
- **without them** — registration succeeded and the first event failed with `NotImplementedException("Did you forget to implement this?")` from the base virtual.

## Changes

`MethodInfo.GetBaseDefinition()` separates the two: an override reports the base-class declaration, a hiding member reports itself.

The same lookup also went through `Type.GetMethod(string)`, which throws `AmbiguousMatchException` as soon as the projection declares any other method of that name — a private `Evolve(string)` helper was enough to break registration outright. Resolution now matches the base declaration's signature.

## Test plan

`HidingDispatchVirtualTests`, all three verified failing without the fix:

- a hiding `Evolve` alongside conventions registers cleanly and dispatch runs the conventional `Apply` (`ACount == 1`, not the hiding method's `+= 100`);
- a same-named helper registers instead of throwing `AmbiguousMatchException`;
- a **real** `override` still owns dispatch over both the evolver and the conventions (`ACount == 100`) — the contract this must not break.

EventTests 749 passed on net9.0 and net10.0; EventStoreTests 114; source generator tests 34.

### DIFF
--- a/src/EventTests/Projections/HidingDispatchVirtualTests.cs
+++ b/src/EventTests/Projections/HidingDispatchVirtualTests.cs
@@ -1,0 +1,96 @@
+using JasperFx.Events;
+using JasperFx.Events.Aggregation;
+using JasperFx.Events.Projections;
+using Shouldly;
+
+namespace EventTests.Projections;
+
+// #656: declaring a dispatch method without `override` hides the base virtual rather than replacing
+// it — the compiler says so with CS0114 — but isOverridden used to count it as an override purely
+// because it is declared outside JasperFx.Events. With conventional methods alongside, registration
+// failed claiming a conflict the author never wrote; without them, dispatch reached the base virtual
+// and threw NotImplementedException at the first event.
+public partial class HidesEvolveAndUsesConventions : SingleStreamProjection<MyAggregate, Guid>
+{
+    public new MyAggregate? Evolve(MyAggregate? snapshot, Guid id, IEvent e)
+    {
+        // Never reached by dispatch — it is not an override. Kept as the author wrote it.
+        snapshot ??= new MyAggregate();
+        snapshot.ACount += 100;
+        return snapshot;
+    }
+
+    public void Apply(AEvent e, MyAggregate a) => a.ACount++;
+}
+
+// A same-named helper used to be enough to break registration outright: Type.GetMethod(string) throws
+// AmbiguousMatchException as soon as more than one method carries the name.
+public partial class OverloadsEvolveName : SingleStreamProjection<MyAggregate, Guid>
+{
+    public string Evolve(string somethingElse) => somethingElse;
+
+    public void Apply(AEvent e, MyAggregate a) => a.ACount++;
+}
+
+public class OverridesEvolveWithConventions : SingleStreamProjection<MyAggregate, Guid>
+{
+    public override MyAggregate? Evolve(MyAggregate? snapshot, Guid id, IEvent e)
+    {
+        snapshot ??= new MyAggregate();
+        snapshot.ACount += 100;
+        return snapshot;
+    }
+}
+
+public class hiding_a_dispatch_virtual
+{
+    [Fact]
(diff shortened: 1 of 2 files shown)


Blackclaws:

Issue #653, opened 2026-08-13
The marten#4787 DI-override emission reintroduces the CS0111 double-load failure #462 removed


Blackclaws:
#462 moved the aggregation dispatcher to a `file`-scoped evolver specifically so that a generator bundled as a built-in analyzer in two referenced packages (e.g. Marten + Polecat) cannot emit the same members twice and fail the consumer build with CS0111.

marten#4787 then added `EmitPartialProjectionWithDIOverride`, which injects an `Evolve` / `EvolveAsync` / `DetermineActionAsync` override *into the user's partial class* — the emission shape #462 removed. Under a double load it fails exactly as before:

```
error CS0111: Type 'AbstractBase' already defines a member called 'Evolve' with the same parameter types
```

## Reproduction

Run two generator instances over one compilation containing a projection that takes the DI-override path — instance conventional methods and no public parameterless constructor, which also covers an abstract projection base class:

```csharp
public abstract partial class AbstractBase : SingleStreamProjection<MyAggregate, Guid>
{
    public void Apply(MyEvent e, MyAggregate agg) { agg.Count++; }
}
```

The two copies must be distinct generator **types**. Two instances of the same type collide on generated file path alone (CS0433), which is a different problem and not what #462 was about:

```csharp
GeneratorDriver driver = CSharpGeneratorDriver.Create(
    new AggregateEvolverGenerator().AsSourceGenerator(),
    new SecondCopyOfTheGenerator().AsSourceGenerator());   // delegates to the same Initialize
```

Every file-scoped emission in the same run is clean; only the injected-override path collides.

## Notes

- Reproduced mechanically in the generator test harness, not observed in the field — I cannot verify from here whether a second shipping package currently bundles this analyzer alongside Marten's copy.
- The tension is real either way: the DI-override path exists precisely because a file-scoped evolver cannot dispatch on a DI-built instance, so this is not simply a matter of reverting it.

PR #660, opened 2026-08-13
fix(source-generator): compose ShouldDelete with Create/Apply for the same event type

### LINKED ISSUE #652: Generated dispatcher does not compile when ShouldDelete and Apply cover the same event type (CS8120)
A projection that declares `ShouldDelete` for an event type the aggregate also has an `Apply` for produces a `DetermineActionAsync` dispatcher with two identical `case` arms, so the generated file does not compile:

```
error CS8120: The switch case is unreachable. It has already been handled by a previous case or it is impossible to match.
```

## Reproduction

```csharp
public class SelfAgg
{
    public Guid Id { get; set; }
    public int Count { get; set; }
    public void Apply(MyEvent e) { Count++; }
}

public partial class PartialOverAggregate : SingleStreamProjection<SelfAgg, Guid>
{
    public bool ShouldDelete(MyEvent e) => false;
}
```

Emitted (trimmed):

```csharp
switch (e.Data)
{
    case global::Test.MyEvent data:
        if (snapshot != null && (_projection.ShouldDelete(data)))
            snapshot = null;
        break;
    case global::Test.MyEvent data:      // <-- CS8120
        snapshot ??= new global::Test.SelfAgg();
        snapshot.Apply(data);
        break;
}
```

`ShouldDelete` and `Apply` are emitted as independent switch arms, so any event type handled by both collides. The intent is visible in the generated code — check the delete predicate, then apply — it just needs to be one arm rather than two.

## Notes

- Reproduces with and without `partial`, so it is independent of the candidacy gate discussed in #650.
- The shape is legitimate: "on this event, delete when the predicate says so, otherwise fold it in".
- The same collision should be checked for `ShouldDelete` + `Create`, and for a `DeleteEvent<T>()` constructor registration naming an event that also has an `Apply`.

Blackclaws:
Closes #652

## Why

An event type covered by both a delete and a `Create`/`Apply` produced two arms in one `DetermineAction` switch. The delete arm's `case T data:` is unguarded, so the second was unreachable and the consumer build failed with CS8120 inside generated code it cannot edit:

```csharp
case global::Test.MyEvent data:          // ShouldDelete arm — unguarded
    if (snapshot != null && (_projection.ShouldDelete(data))) snapshot = null;
    break;
case global::Test.MyEvent data:          // CS8120: unreachable
    snapshot ??= new global::Test.SelfAgg();
    snapshot.Apply(data);
    break;
```

The shape is legitimate — "delete on this event when the predicate says so, otherwise fold it in" — and reproduces with or without `partial`, so it is independent of the candidacy gate in #650.

## Changes

The two projection dispatch shapes — the file-scoped evolver and the override injected into the user's class — carried near-identical copies of this arm emission, so it is now one `EmitDetermineActionSwitchArms` and the fix lands in both.

Delete arms own their event type. A `ShouldDelete` predicate is just that: when it returns `false` the event still has to be folded in, so the `Create`/`Apply` body moves into the arm's `else`:

```csharp
case global::Test.MyEvent data:
    if (snapshot != null && (_projection.ShouldDelete(data))) { snapshot = null; }
    else { snapshot ??= new global::Test.SelfAgg(); snapshot.Apply(data); }
    break;
```

A constructor-registered `DeleteEvent<T>()` is unconditional — mirroring the runtime's `MatchesAnyDeleteType` short circuit in `buildActionAsync` — so those event types drop their `Create`/`Apply` entirely. That also closes a second CS8120 collision that was never filtered: ctor-delete types were only checked against `ShouldDelete` types, never against `Create`/`Apply`.

**Behaviour change worth review.** The self-aggregating dispatcher avoided the collision by skipping the `Create`/`Apply` arm for any type with a `ShouldDelete`, which silently disabled the Apply instead — `ShouldDelete(E) => false` meant `Apply(E)` never ran. It now composes the two the same way as the projection paths.

## Test plan

`ShouldDeleteCompositionTests`: delete + apply compiles with one arm; a false predicate still applies the event; a ctor `DeleteEvent<T>()` suppresses the apply; delete + create compiles.

Source generator tests 38 passed; EventTests 746 on net9.0 and net10.0; EventStoreTests 114.

## Merge order

Touches `EmitDetermineActionAsOverride`, which #653 deletes — **this one before #653** is the cheaper order.

### DIFF
--- a/src/JasperFx.Events.SourceGenerator.Tests/GeneratorHarness.cs
+++ b/src/JasperFx.Events.SourceGenerator.Tests/GeneratorHarness.cs
@@ -1,0 +1,110 @@
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace JasperFx.Events.SourceGenerator.Tests;
+
+/// <summary>
+/// Shared compilation harness for the generator's regression suites.
+///
+/// <para>Each issue's tests live in their own file rather than all landing at the end of
+/// AggregateEvolverGeneratorTests, so independent fixes to the generator do not collide on one test
+/// file. This type is what makes that practical — the content is deliberately identical wherever it
+/// appears, so branches that both introduce it merge without a conflict.</para>
+/// </summary>
+internal static class GeneratorHarness
+{
+    private static List<MetadataReference> References()
+    {
+        var references = new List<MetadataReference>
+        {
+            MetadataReference.CreateFromFile(typeof(object).Assembly.Location),
+            MetadataReference.CreateFromFile(typeof(IEvent).Assembly.Location),
+            MetadataReference.CreateFromFile(typeof(Guid).Assembly.Location),
+        };
+
+        var runtimeDir = System.IO.Path.GetDirectoryName(typeof(object).Assembly.Location)!;
+        references.Add(MetadataReference.CreateFromFile(System.IO.Path.Combine(runtimeDir, "System.Runtime.dll")));
+        references.Add(MetadataReference.CreateFromFile(System.IO.Path.Combine(runtimeDir, "System.Collections.dll")));
+
+        return references;
+    }
+
+    private static CSharpCompilation Compilation(string source)
+    {
+        return CSharpCompilation.Create(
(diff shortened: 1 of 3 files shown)


Blackclaws:

PR #661, opened 2026-08-13
fix(source-generator): hand the evolver the registered projection instead of a shadow

### LINKED ISSUE #653: The marten#4787 DI-override emission reintroduces the CS0111 double-load failure #462 removed

Blackclaws:
Closes #653

## Why

An evolver whose conventional methods live on the projection instance had to get an instance from somewhere, and it built its own: `new TProjection()`, or `RuntimeHelpers.GetUninitializedObject` when there was no public parameterless constructor. Both are shadows of the projection the store actually registered, so injected dependencies were null (marten#4787) and anything the real constructor set was missing.

marten#4787 addressed the DI half by emitting the dispatcher as an override injected into the user's partial class — the member-injection shape #462 had just removed — and with it came back the CS0111 failure when the generator is bundled in two referenced packages and loads twice.

## Changes

Remove the need for member injection rather than patch it. The evolver declares a constructor taking the projection type, and the runtime passes `this`:

```csharp
private object activateEvolver(Type evolverType)
{
    var boundConstructor = evolverType.GetConstructors()
        .FirstOrDefault(c => c.GetParameters().Length == 1
                             && c.GetParameters()[0].ParameterType.IsInstanceOfType(this));

    return boundConstructor != null ? boundConstructor.Invoke([this]) : Activator.CreateInstance(evolverType)!;
}
```

Dispatch runs on the registered instance in every case, the dispatcher stays a file-scoped type two loaded copies cannot collide over, and the injected-override emission — along with its `Evolve`/`EvolveAsync`/`DetermineActionAsync` override emitters and the `dispatchOnThis` plumbing — is gone. Evolvers that need nothing from the projection keep their parameterless constructor, which is also the fallback for evolvers generated before this change.

## Version coupling

The generator and the `JasperFx.Events` runtime now have to move together: an evolver with only the projection constructor cannot be activated by a runtime without `activateEvolver`. They ship as a matched pair (Marten bundles the analyzer matching its `JasperFx.Events` pin), but it is a coupling worth stating explicitly.

## Test plan

- `ProjectionInstanceBindingTests` (runtime): a DI-built projection's injected dependency is reachable during dispatch; and `a_projection_with_a_parameterless_constructor_dispatches_on_itself_too` **fails on `main`** — the shadow absorbs the mutation — and passes here. Verified both directions. The DI test alone would not discriminate, since member injection also dispatched on the real instance.
- `EvolverInstanceBindingTests` (generator): the DI shape emits a file-scoped evolver taking the projection through its constructor with no shadow and nothing written into the user's class; and it survives the generator being loaded twice.
- Source generator tests 35 passed; EventTests 748 on net9.0 and net10.0; EventStoreTests 114.
- Marten 9.23.0's `src/EventSourcingTests` builds clean with this generator swapped in for the published analyzer.

## Merge order

This deletes `EmitPartialProjectionWithDIOverride` and its three override emitters. #650/#651 routes generic and non-visible projections *to* that emitter, and #652 refactors one of the three — so this should land **last**, and rebases to keep the emitter for the naming-constrained shapes only if #651 lands first.

### DIFF
--- a/docs/codegen/projection-apply-dispatch.md
+++ b/docs/codegen/projection-apply-dispatch.md
@@ -31,7 +31,7 @@ The SG emits one of five shapes through `EvolverCodeEmitter`
 
 | Mode | Emits | How runtime picks it up |
 |---|---|---|
-| `PartialProjection` | Partial method on the user's projection class (`Evolve` / `EvolveAsync` / `DetermineActionAsync`) | `JasperFxAggregationProjectionBase.isOverridden(...)` returns `true` for the generated override and `_usesConventionalApplication = false` — `AggregateApplication` is never invoked for dispatch. |
+| `PartialProjection` | Standalone `file`-scoped evolver + assembly-level `[GeneratedEvolverAttribute]` carrying the projection type. When its conventional methods live on the projection instance, the evolver takes that projection through its constructor. | `tryUseAssemblyRegisteredEvolver(...)` selects the registration and `activateEvolver` builds the evolver, passing `this` when it asks for the projection, so dispatch runs on the registered instance. `_usesConventionalApplication = false` — `AggregateApplication` is never invoked for dispatch. |
 | `SelfAggregating` | Standalone evolver class implementing `IGeneratedSyncEvolver<TDoc,TId>` or `IGeneratedSyncDetermineAction<TDoc,TId>` + an assembly-level `[GeneratedEvolverAttribute]` | `JasperFxAggregationProjectionBase.tryUseAssemblyRegisteredEvolver(...)` activates and binds the evolver. |
 | `SelfAggregatingEvolve` | Standalone evolver class implementing `IGeneratedAsyncEvolver<TDoc,TId>` + `[GeneratedEvolverAttribute]` | Same as above. |
 | `EventProjection` | Partial `ApplyAsync(TOperations, IEvent, CancellationToken)` override on the user's `JasperFxEventProjectionBase<TOperations>` subclass | The override wins at vtable dispatch. `EventProjectionApplication` is only used when the base `ApplyAsync` is not overridden. |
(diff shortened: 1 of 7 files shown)


Blackclaws:

PR #662, opened 2026-08-13
Projection dispatch: narrow the `partial` requirement, bind the real projection, and fix four adjacent defects

### LINKED ISSUE #650: Projection subclasses are silently skipped unless `partial`, though only the injected-dispatcher paths need it

Blackclaws:
Rolls up the six projection-dispatch fixes into one branch. Every commit from the individual PRs is preserved and each is brought in with a merge commit, so merging this closes all six as merged rather than leaving them to be re-applied.

Closes #650, #652, #653, #654, #655, #656

| PR | issue | what it fixes |
|---|---|---|
| #651 | #650 | projection subclasses silently skipped unless `partial` |
| #657 | #655 | opt-out for lambda registration APIs that no longer exist |
| #658 | #654 | non-partial EventProjection loses published types with no diagnostic |
| #659 | #656 | a method that *hides* `Evolve` is treated as an override |
| #660 | #652 | CS8120 when `ShouldDelete` and `Create`/`Apply` cover the same event type |
| #661 | #653 | the evolver dispatches through a shadow instead of the registered projection |

Each PR stands on its own and can be reviewed there; this branch exists because two of them overlap semantically and the resolution is worth seeing as one thing.

## The integration decisions

Four of the six merged with no conflict or with a one-file resolution. The `#651 + #661` and `#660 + #661` overlaps are real: #661 removes the member-injection emission that #651 routes to and #660 refactors. Resolved as follows, and recorded in the merge commit:

**Member injection survives only for what a separate file cannot name** — a generic projection, or one behind `private`/`protected`. #661's own reason for keeping it disappears: the evolver now takes the projection through its constructor and the runtime hands it the registered instance, so the marten#4787 shape (instance conventional methods, no public parameterless constructor) keeps the file-scoped evolver like everything else. `RequiresMemberInjection` reduces to the naming test, and `BuildProjectionInstanceExpression` / `HasPublicParameterlessCtor` go with the shadow they served.

**The `dispatchOnThis` plumbing #661 deleted is restored**, because the injected-override emitters it feeds are still reachable for those shapes — and #660's shared `EmitDetermineActionSwitchArms` serves both emission paths, so the CS8120 fix applies to each.

**The combined result is better than either PR alone.** A non-partial DI-activated projection now needs no `partial` at all, where #651 on its own still required it. `PartialRequirementTests` carries that as `non_partial_di_activated_projection_dispatches_through_the_registered_instance`, replacing the error assertion that was only correct while #651 stood alone. `JFXEVT003` is left with one reason instead of two.

## Where `partial` ends up

After all six, the modifier is required in exactly two places, both reported as `JFXEVT003` at build time:

- an `EventProjection` subclass using conventional methods — its dispatcher *is* an `ApplyAsync` override on the user's class;
- a generic projection, or one behind `private`/`protected`, which a file-scoped type cannot name — and then every containing type must be `partial` too.

Everything else — including DI-activated projections and every ordinary `SingleStreamProjection` / `MultiStreamProjection` subclass — is dispatched by a standalone file-scoped evolver that needs nothing from the declaration.

## Test plan

Full solution, both target frameworks:

- `JasperFx.Events.SourceGenerator.Tests` 56
- `EventTests` 751 (net9.0 and net10.0)
- `EventStoreTests` 114
- `CommandLineTests` 295, `CoreTests` 574, `CodegenTests` 419

Each PR's regression tests live in their own file sharing a `GeneratorHarness`, so the roll-up carries all of them without any test-file merge resolution.

## Note on CI

`CommandLineTests.CommandExecutorTester.execute_single_command_synchronously` fails intermittently on net9.0 in this repo, unrelated to any of these changes — it fails on #659 and #660, which touch the runtime and the generator respectively and neither of which goes near `CommandLineTests`, while the same test passes on net10.0 in the same run and 295/295 pass locally. The test's own comment documents the flake from #587 / #589: it assumes the "Searching '&lt;assembly&gt;' for commands" banner is always a prefix of the captured output, and CI shows it arriving after the command output instead. Worth its own fix.


Blackclaws:
Individual PRs: #651, #657, #658, #659, #660, #661 — all six are merged into this branch as merge commits, so merging here closes each of them.

PR #668, opened 2026-08-16
gh-667: stop CalculateCeiling claiming a bound it never scanned

### LINKED ISSUE #667: EventPage.CalculateCeiling can claim an unscanned bound when a whole batch is skipped
Split out of [marten#5239](https://github.com/JasperFx/marten/issues/5239) / [marten#5242](https://github.com/JasperFx/marten/pull/5242), where it was deliberately left unfixed rather than worked around locally.

`EventPage.CalculateCeiling`:

```csharp
public void CalculateCeiling(int batchSize, long highWaterMark, int skippedEvents = 0)
{
    Ceiling = ((base.Count + skippedEvents == batchSize && base.Count > 0) ? this.Last().Sequence : highWaterMark);
}
```

The `base.Count > 0` conjunct means a page whose rows were **all** skipped — every one an unknown event type or a deserialization failure — takes the `highWaterMark` branch even though the query's `LIMIT` was saturated and rows beyond it inside the same range were never seen. The consumer writes `Ceiling` as durable projection progress, so those unseen events are skipped permanently.

Marten's `loadNormalAsync` reaches this with `highWaterMark = request.HighWater`, so the ceiling jumps to the full high-water mark. marten#5242 fixed a much more common instance of the same shape in the window-step loader (which was passing `highWater` for a query bounded by a 10,000-wide window), and while doing so tightened the window-step advance to key off rows read rather than `page.Count`, precisely so an all-skipped window could not step over `LIMIT`-truncated events. That local guard cannot help `loadNormalAsync`, because the over-claim happens inside this method.

## Why it is not obviously a bug

Guaranteeing forward progress is presumably the intent: if a batch is entirely undeserializable and the ceiling stayed at the floor, the shard would re-read the same poison rows forever. Trading correctness for liveness there is defensible.

But the trade is currently silent and unconditional. Two things seem worth separating:

- **`Count == 0` because the range genuinely held nothing** — `highWaterMark` is correct, the query exhausted the range.
- **`Count == 0` because every one of `batchSize` rows was skipped** — the query did *not* exhaust the range; it stopped at the `LIMIT`. Claiming `highWaterMark` here is the same defect the `Count + skipped == batchSize` arm exists to prevent, just with the `Count > 0` guard steering it into the wrong branch.

The second case is distinguishable: `skippedEvents == batchSize`. It could take the `this.Last().Sequence` arm if the page retained the last *observed* sequence rather than the last successfully-deserialized one, which would keep liveness (progress still advances past the poison rows) without over-claiming.

## Practical impact

Narrow — it needs an entire batch of consecutive events that all fail to deserialize, which usually means a removed or renamed event type in a contiguous span. Rare, but the failure mode is the same as marten#5239: no exception, no dead-letter beyond the individual skips, a shard reporting itself caught up, and a read model permanently missing data.

Filing rather than patching downstream because every store consuming `IEventLoader` inherits the behaviour, and a Marten-local divergence here would be worse than the gap.

jeremydmiller:
Closes #667.

## The defect

```csharp
Ceiling = (Count + skippedEvents == batchSize && Count > 0) ? this.Last().Sequence : highWaterMark;
```

The `Count > 0` conjunct was added for [marten#4663](https://github.com/JasperFx/marten/issues/4663), to keep `Last()` off an empty page. It also steers a page whose rows were **all** skipped into the `highWaterMark` branch — and when the batch was saturated that is wrong twice over. The query stopped at its `LIMIT` rather than exhausting the range, so rows past it inside the same range were never read, and `Ceiling` is written as durable projection progress (`EventRange` → `MarkSuccessAsync`, and `LastEnqueued`). Those events are skipped permanently: no exception, no dead letter beyond the individual skips, a shard reporting itself caught up over a read model missing data.

## The change

The predicate is now saturation alone, and the ceiling is the last row **observed** rather than the last one kept:

```csharp
var saturated = batchSize > 0 && Count + skippedEvents == batchSize;
if (!saturated) { Ceiling = highWaterMark; return; }

var lastKept = Count > 0 ? this[Count - 1].Sequence : 0;
var lastObserved = Math.Max(lastKept, LastSkippedSequence);
Ceiling = lastObserved > 0 ? lastObserved : highWaterMark;
```

Skips are observations. They move the ceiling forward — so a poison span is stepped over exactly once instead of re-read forever, which is the liveness marten#4663 was protecting — but only as far as the loader actually got.

Rows arrive in sequence order, so `max(lastKept, lastSkipped)` is the final row under the `LIMIT`. Taking the max also fixes a smaller waste on the common path: a page ending in skipped rows previously stopped at the last *kept* sequence and re-read that tail on the next batch.

## How a loader reports it

A page cannot invent a sequence it was never told, so this adds `RecordSkippedEvent`. The overload worth knowing about takes `IEventFailureContext`:

```csharp
catch (UnknownEventTypeException e) when (request.ErrorOptions.SkipUnknownEvents)
{
    page.RecordSkippedEvent(e);
    skippedEvents++;
}
```

That interface already guarantees `Sequence` and every skippable read failure in the stores implements it, so the sentinel handling — "the sequence could not be determined", which Marten spells `-1` — stays in one place rather than being re-derived per exception type. A non-positive sequence is ignored, so the sentinel degrades to the fallback instead of pinning the ceiling below the floor.

## Ships inert until stores adopt it

A loader that reports counts but not sequences still lands on `highWaterMark` for an all-skipped saturated page. That is the pre-fix behavior kept deliberately: stalling at the floor instead would reintroduce marten#4663's infinite poison loop, and the page has nothing better to go on.

Marten, Polecat and Fisher all pass skip counts today and none report sequences, so **none of them changes behavior on this bump** — each closes the gap by adding `RecordSkippedEvent` to its loader's catch blocks. Marten's two skip exceptions already carry the sequence; the window-step loader in [marten#5242](https://github.com/JasperFx/marten/pull/5242) is the other site.

## Tests

`EventPageTests` goes from 3 to 13, covering saturated-and-all-skipped (the defect), partial-and-all-skipped (which correctly still claims the high water mark — this is why the fix keys off saturation rather than off the page being empty), the no-sequence fallback, trailing and leading skips against a non-empty page, the `-1` sentinel, and the failure-context overload.

Full `EventTests` 763/0 and `EventStoreTests` 121/0.

### DIFF
--- a/src/EventTests/Projections/EventPageTests.cs
+++ b/src/EventTests/Projections/EventPageTests.cs
@@ -1,4 +1,5 @@
 using JasperFx.Events;
+using JasperFx.Events.Daemon;
 using JasperFx.Events.Projections;
 using Shouldly;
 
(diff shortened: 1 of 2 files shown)


jeremydmiller:
Closing in favour of #670, which fixes the same defect and got there first with a better core.

Two things it does better than this PR:

- **`LastObservedSequence`** — "last row read, kept or skipped" is the honest abstraction, and it matches the issue's own vocabulary. This PR tracked skips specifically (`LastSkippedSequence`), which is narrower for no gain.
- **Sourcing the sequence from the row rather than the exception.** This PR's headline overload took `IEventFailureContext`, which reads nicely in a `catch` block and is wrong exactly where it matters: Marten's `UnknownEventTypeException` carries `Sequence = -1` when the stored .NET type no longer resolves, which is the removed-or-renamed event type that fills a whole batch with skips in the first place.

The one thing worth keeping from here — a guard so a reported sequence at or below the floor can't drive durable progress backwards — is now [pushed onto #670](https://github.com/JasperFx/jasperfx/pull/670#issuecomment-5315069942) as a follow-up commit, along with the three tests that pin it.

Not carried over, deliberately: this PR also used `max(lastKept, lastObserved)` on the partly-skipped branch, which avoids re-reading skips that trail the last kept event. It's correct, but it changes behaviour on a path that isn't broken, and it doesn't belong bundled into a correctness fix. Filing separately if it ever earns its own justification.

PR #670, opened 2026-08-17
Do not claim an unscanned ceiling when a whole batch is skipped

### LINKED ISSUE #667: EventPage.CalculateCeiling can claim an unscanned bound when a whole batch is skipped

arnelirobles:
Fixes #667.

## The change

`CalculateCeiling` no longer claims the caller's high water mark when a full batch was entirely skipped. It uses the last sequence the loader actually read, when the loader reports one.

The over-claim mattered because the consumer writes `Ceiling` as durable projection progress. An all-skipped page still saturated its `LIMIT`, so rows beyond it inside the same range were never looked at, and jumping to `highWaterMark` skipped them permanently: no exception, no dead-letter beyond the individual skips, and a shard reporting itself caught up.

## Keeping the liveness the fallback exists for

The `highWaterMark` fallback was added by marten#4663 so a batch of undeserializable rows could not leave a shard re-reading the same poison forever. That property is kept: `LastObservedSequence` is past those rows by definition, so progress still advances. What it stops doing is advancing past rows nobody read.

The one thing the page could not do before was know that sequence, because an `EventPage` only ever contains events that deserialized. So it has to be told, which is the new property:

```csharp
public long? LastObservedSequence { get; set; }
```

## Nothing changes for existing callers

`LastObservedSequence` defaults to null and the ceiling falls back to `highWaterMark` exactly as before, so this is opt-in per loader and no existing behaviour moves. The existing `calculate_ceiling_when_full_batch_was_entirely_skipped_does_not_throw` test, which pins the marten#4663 fallback, passes unchanged and was deliberately left alone.

The method now returns early per branch rather than nesting a conditional, which was needed to comment the three cases separately. No behavioural change in the two existing branches.

Marten's `loadNormalAsync` is where this bites in practice, and it would need to set the property as it reads rows. That is a Marten-side change and is not in this PR.

## Tests

758 pass in `EventTests` on net10.0. Four added:

| Test | Pins |
|---|---|
| `an_entirely_skipped_full_batch_claims_only_what_the_query_read` | the fix: ceiling is 10, not 1000 |
| `an_entirely_skipped_full_batch_moves_past_the_poison_rows_without_passing_the_query` | liveness and the fix together |
| `a_partly_skipped_full_batch_still_uses_its_last_surviving_event` | the new property must not take over while the page holds events |
| `an_unfilled_batch_ignores_the_last_observed_sequence` | an unfilled batch genuinely exhausted its range, so the high water mark still wins |

**Proven by mutation.** Reverting the new branch to `Ceiling = highWaterMark` fails the first two and leaves the other two green, which is correct: those two are guards on paths the fix does not touch.

Worth mentioning because it nearly went in wrong: the liveness test first asserted only `Ceiling.ShouldBeGreaterThan(Floor)`, and it **passed against the bug**, since the old ceiling of 1000 also clears a floor of 5. It now asserts the exact value as well.

## Note on net9.0

Only the net10.0 leg ran locally; the .NET 9 runtime is not installed on this machine. The change is a branch in one method with no framework-conditional code, so CI covering both should be enough, but I have not run net9.0 myself and did not want to imply otherwise.

### DIFF
--- a/src/EventTests/Projections/EventPageTests.cs
+++ b/src/EventTests/Projections/EventPageTests.cs
@@ -45,4 +45,67 @@ public class EventPageTests
 
         page.Ceiling.ShouldBe(1000);
     }
+
+    [Fact]
+    public void an_entirely_skipped_full_batch_claims_only_what_the_query_read()
+    {
+        // https://github.com/JasperFx/jasperfx/issues/667. Ten rows read, ten skipped, so the
+        // batch was saturated and the LIMIT stopped the query at sequence 10. Everything between
+        // 11 and the high water mark was never looked at, and the consumer writes Ceiling as
+        // durable projection progress, so claiming 1000 here skips those events permanently.
+        var page = new EventPage(0) { LastObservedSequence = 10 };
+
+        page.CalculateCeiling(10, 1000, skippedEvents: 10);
+
+        page.Ceiling.ShouldBe(10);
+    }
+
+    [Fact]
+    public void an_entirely_skipped_full_batch_moves_past_the_poison_rows_without_passing_the_query()
+    {
+        // The reason the high water mark fallback exists at all is liveness: a ceiling stuck at
+        // the floor leaves the shard re-reading the same undeserializable rows forever. The fix
+        // has to keep that while dropping the over-claim, so both bounds are asserted together.
+        //
+        // Written with an exact value rather than only ShouldBeGreaterThan(Floor). A range
+        // assertion here would pass against the very behaviour this pins, since the old ceiling of
+        // 1000 is also greater than the floor, and a test that passes against the bug it describes
+        // is not a test.
+        var page = new EventPage(floor: 5) { LastObservedSequence = 15 };
+
+        page.CalculateCeiling(10, 1000, skippedEvents: 10);
+
+        page.Ceiling.ShouldBe(15);
+        page.Ceiling.ShouldBeGreaterThan(page.Floor);
+    }
+
+    [Fact]
+    public void a_partly_skipped_full_batch_still_uses_its_last_surviving_event()
+    {
+        // LastObservedSequence must not take over whenever it happens to be set. While the page
+        // holds anything, the last event in it is the ceiling, exactly as before: those rows were
+        // read and kept, and the ones after them inside the batch were read and skipped.
+        var page = new EventPage(0)
+        {
+            new Event<AEvent>(new AEvent()) { Sequence = 4 }
+        };
+        page.LastObservedSequence = 9;
+
+        page.CalculateCeiling(10, 1000, skippedEvents: 9);
+
+        page.Ceiling.ShouldBe(4);
+    }
+
+    [Fact]
+    public void an_unfilled_batch_ignores_the_last_observed_sequence()
+    {
+        // A batch that did not fill genuinely exhausted its range, so the high water mark is
+        // correct and must win even when the loader reported what it last read. Without this,
+        // a loader that always sets the property would stop advancing past sparse ranges.
+        var page = new EventPage(0) { LastObservedSequence = 7 };
+
+        page.CalculateCeiling(10, 1000, skippedEvents: 2);
+
+        page.Ceiling.ShouldBe(1000);
+    }
 }
--- a/src/EventTests/Projections/EventPageTests.cs
+++ b/src/EventTests/Projections/EventPageTests.cs
@@ -108,4 +108,44 @@ public class EventPageTests
 
         page.Ceiling.ShouldBe(1000);
     }
+
+    [Fact]
+    public void an_undetermined_sequence_sentinel_is_not_treated_as_an_observation()
+    {
+        // Marten's UnknownEventTypeException reports -1 when the throw site had no row in hand, and
+        // the read path hits exactly that when an event's stored .NET type no longer resolves -- the
+        // removed-or-renamed event type that fills a whole batch with skips in the first place. A
+        // loader that reported the exception's sequence rather than the row's would land here.
+        // Acting on it would set the ceiling below the floor and march durable progress backwards.
+        var page = new EventPage(100) { LastObservedSequence = -1 };
+
+        page.CalculateCeiling(10, 1000, skippedEvents: 10);
+
+        page.Ceiling.ShouldBe(1000);
+    }
+
+    [Fact]
+    public void an_observed_sequence_at_or_below_the_floor_is_not_trusted()
+    {
+        // Every row this query could have read is above Floor, so a value down here is stale -- left
+        // over from an earlier range rather than observed in this one. Same reasoning as the
+        // sentinel: fall back rather than move progress backwards.
+        var page = new EventPage(100) { LastObservedSequence = 100 };
+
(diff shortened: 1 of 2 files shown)


jeremydmiller:
Thanks for this — taking your approach over my own. I had #668 open against the same issue and I'm closing it in favour of this one.

Two things yours does better, worth stating since they're the reason:

- **`LastObservedSequence` is the right name and the right concept.** Mine tracked skips specifically (`LastSkippedSequence`); "last row read, kept or skipped" is the honest abstraction and it matches the vocabulary in the issue.
- **Sourcing it from the row rather than the exception.** I'd hung my primary overload off `IEventFailureContext`, which reads well at a call site and is wrong where it counts — see below.

I've pushed one commit onto this branch rather than asking for a change, since it's small and mechanical.

## The guard

`Ceiling = LastObservedSequence ?? highWaterMark` acts on whatever the loader reported, and a loader cannot always report something usable.

Marten's `UnknownEventTypeException` carries `Sequence = -1` when the throw site had no row in hand. That isn't a corner: `EventDocumentStorage.eventMappingForDotNetTypeName` calls `Events.TypeForDotNetName(dotnetTypeName)`, and when the stored .NET type no longer resolves, `EventGraph.TypeForDotNetName` throws through the single-argument constructor — the `catch` there only catches `ArgumentNullException`, so it escapes with the sentinel. A removed or renamed event type is *the* scenario that fills a whole batch with skips, so a loader adopting this by reporting `e.Sequence` would set `Ceiling = -1` and move durable progress backwards, past a floor the shard already committed. That's worse than the over-claim being fixed.

The guard is `Floor`, not zero, because `Floor` is what makes it true — every row the query could have read is above it, so a value at or below it wasn't observed in this range, whether it's a sentinel or a leftover from an earlier one:

```csharp
Ceiling = LastObservedSequence > Floor ? LastObservedSequence.Value : highWaterMark;
```

Falling back leaves the pre-#667 behaviour for that page, which is the same call the un-adopted path already makes.

I also documented on the property that a loader should take the sequence from the reader's own sequence column. The exception is missing it precisely when it's most needed, and a discarded value is a missed fix rather than a fixed bug.

## Tests

Three added: the sentinel, a stale value sitting exactly at the floor, and the first legitimate sequence above it so the guard can't quietly become an off-by-one. Mutation-checked the way you did yours — reverting to `?? highWaterMark` fails the first two and leaves the boundary test green, which is correct since that path is unchanged.

761 in `EventTests` on net10.0 (your 758 plus 3). I didn't run net9.0 either; CI covers both.

## Left alone deliberately

Your `Count > 0` branch. Mine used `max(lastKept, lastObserved)`, which also advances past skips trailing the last kept event and saves re-reading that tail — but it's a behaviour change on a path that isn't broken, and your call to leave it untouched is the more conservative one. Worth a separate issue if the re-read ever shows up in practice; it shouldn't be bundled into a correctness fix.

Also noting for anyone reading later: this ships inert. Marten, Polecat and Fisher all pass skip counts and none set `LastObservedSequence` yet, so nothing changes behaviour on the bump until each loader adopts it.

arnelirobles:
Thanks, and the guard is a straight catch. `?? highWaterMark` only ever handled null, so I had guarded the case where a loader says nothing and missed the case where it says something wrong, which is the one that actually shows up.

The `Sequence = -1` path is the part I should have followed and did not. A removed or renamed event type is exactly what fills a batch with skips, so the first real adopter would have hit it, and moving durable progress *backwards* past a committed floor is worse than the over-claim I opened the PR about. Trading a silent skip for a silent rewind is not a fix.

`> Floor` is also the better test than a null-or-negative check, for the reason you gave rather than for being stricter: every row the query could have read is above the floor, so anything at or below it was not observed in this range whatever the reason. That covers the sentinel and a stale value from an earlier page with one condition, and it stays true if some other loader invents a third way to report nothing useful.

Agreed on documenting that the sequence comes from the reader's own column rather than the exception. That was the half of my own design I got right and did not write down, and a discarded value is worse than an obvious gap because nothing looks wrong.

Also agreed on leaving `Count > 0` alone. Your `max(lastKept, lastObserved)` is better behaviour and I would rather it arrived as its own change with its own reasoning than rode in on a correctness fix, since the two would be impossible to separate later if the re-read ever mattered.

Happy to take the Marten-side adoption as a follow-up now that this ships inert, unless you would rather do it with the loader work you already have in flight. `loadNormalAsync` is the one that reaches this with `request.HighWater`, so it is the one where the fix actually changes anything.

jeremydmiller:
I'm having to push Marten for something unrelated today, so that's already brewing now.

Issue #669, opened 2026-08-17
No Events accessor on the JasperFx.Events.Documents session contracts — a consumer-opened session cannot reach the event store


jeremydmiller:
`JasperFx.Events.Documents` has no route from a session to its event store, so a consumer that opens its own session cannot touch events store-agnostically.

## The gap

`IDocumentSessionFactory` hands back the shared session contracts:

```csharp
IDocumentSessionOperations LightweightSession();
IDocumentReadOperations    QuerySession();
```

Neither returned contract exposes an `Events` accessor. Marten declares
`new IEventStoreOperations Events { get; }` on **its own** `IDocumentSession`, not on the
JasperFx contract — and Polecat and Fisher do the same on theirs. So the moment a consumer
takes the store-agnostic route, the event store becomes unreachable from the session it
just opened.

This is only about the **accessor**. The contracts on the far side of it already exist and
are already sufficient:

| what the consumer needs | contract that already declares it |
|---|---|
| `Append` / `StartStream` / `FetchForWriting` | `IEventStoreOperations` |
| `FetchStreamStateAsync` | `IQueryEventStore` / `IReadOnlyEventStore` |

## Why it doesn't show up in the handler case

It doesn't bite chain parameters at all, because wolverine#3956 (6.28.2) lets a handler take
`IEventStoreOperations` directly and Wolverine fills it from the chain's own session. We have
converted a good number of handlers on exactly that basis and it works in all three stores.

It bites the two shapes where the consumer, not Wolverine, decides when a session exists:

1. **Background / timer publishers.** An evaluator that wakes on a timer, opens a session and
   appends. It never had a chain session to bind to, so `IDocumentSessionFactory` is the right
   contract for the session — and then there is no way to append.

2. **A deliberate second read session.** Our hottest batch handler opens a query session
   *separately from* the chain's writing session on purpose (its projections' `Evolve` mutates
   its input, so the decision copies must not be tracked by the session that is about to
   commit). That session needs one narrow `Events.FetchStreamStateAsync` probe to validate a
   per-service cache. Moving just that probe onto the chain's session would split the group's
   reads across two connections and let the version probe and the document loads disagree —
   which is precisely the staleness the cache exists to detect, so it is not a workaround we
   are willing to take on that path.

## Proposed shape

Mirror on the contracts what all three products already declare on their own session types:

```csharp
public interface IDocumentSessionOperations : IDocumentWriteOperations
{
    IEventStoreOperations Events { get; }
[...]

Issue #665, opened 2026-08-15
IDocumentReadOperations.LoadAsync<T> is Guid|string only — store-agnostic code cannot do a strong-typed by-id load


jeremydmiller:
Follow-on from #647, from the first real consumer of `JasperFx.Events.Documents`.

## Context

CritterWatch is decomposing into a store-agnostic core plus thin Marten / Polecat / Fisher bindings. The core project references no store at all — that absence is enforced by the compiler and is the whole point of the design — so every document operation goes through `JasperFx.Events.Documents`. That has worked well; this is the first place the surface is too narrow to express something the stores can already do.

## The gap

`IDocumentReadOperations.LoadAsync<T>` offers `Guid` and `string`. A document identified by a **strong-typed id** cannot be loaded by id at all:

```csharp
// AlertRecord's identity is `public record struct AlertId(string Id)`, registered
// with StoreOptions.RegisterValueType<AlertId>().
return await session.LoadAsync<AlertRecord>(new AlertId(id), ct);
// error CS1503: cannot convert from 'AlertId' to 'System.Guid'
```

Passing the wrapped `string` is not an alternative: it compiles, then throws `DocumentIdTypeMismatchException` at runtime, because the document's id is the wrapper and not its primitive.

So store-agnostic code has to fall back to a LINQ query on the identity:

```csharp
var alertId = new AlertId(id);
return await session.Query<AlertRecord>().FirstOrDefaultAsync(x => x.Id == alertId, ct);
```

which works and still resolves on the primary key, but is a query standing in for a load.

## On scope

#647 explicitly scoped the surface tight — "resist growing this" — and deliberately omitted `SingleOrDefaultAsync`, `SumAsync`, `MinAsync`, `MaxAsync`, `LongCountAsync`. That was the right call and this is not a request to relitigate it.

The distinction I would draw: those omissions all have a straightforward expression through what is already there. This one does not — there is **no** way to load a strong-typed-id document by id through the abstraction, and strong-typed identifiers are a first-class, documented feature in Marten, Polecat and Fisher alike. The abstraction can currently express by-id loads only for documents that happen to use a primitive identity, which is a capability cliff rather than a convenience gap.

An `object`-typed overload (what Marten already offers) would be the minimum, since the stores each resolve the value type from their own registration.

## Related

[...]

PR #675, opened 2026-08-17
gh-673: expose a session's pending stream actions on the document contract

### LINKED ISSUE #673: Expose the session's pending stream actions on the JasperFx.Events.Documents contract
Add a pending-stream-actions accessor to the `JasperFx.Events.Documents` session contract, so a consumer can read the `StreamAction`s a session has queued but not yet committed without naming a store.

## Why it can't be done today

All three stores surface the *same* `JasperFx.Events.StreamAction` collection, from three different places:

| store | spelling |
|---|---|
| Marten | `PendingChanges.Streams()` |
| Polecat | `PendingChanges.Streams` |
| Fisher | `Events.PendingStreams` (no `PendingChanges` facade at all) |

The payload type is already shared — only the accessor differs. This is a naming gap, not a capability gap.

Grep over `JasperFx.Events` confirms no `PendingChanges` / `PendingStreams` / `IChangeSet` abstraction exists today.

## Consumer motivation

CritterWatch compiles one shared source tree into three store flavours. `PendingAlertResolutions` is one of eight files it has to maintain in triplicate purely because of the three spellings above, and it is what blocks `ProjectionAlertEvaluator` from moving into the store-agnostic core.

The divergence is recorded in CritterWatch at `src/CritterWatch.Services.Fisher/PendingAlertResolutions.Fisher.cs:7-11`.

## Acceptance

- A member on `IDocumentSessionOperations` (or a sibling contract) returning the queued `IReadOnlyList<StreamAction>`.
- Implemented explicitly on Marten, Polecat and Fisher.

## ⚠️ One design constraint

**If this ships as a default interface member, it must throw, not return empty.**

A silently-empty default means a consumer's alert resolutions (or equivalent) are silently dropped, with a clean build and green tests. This is the same hazard class as `IDocumentReadOperations.Events` before #669 and `LoadAsync<T>(object)` — both of which correctly chose a throwing default. Please match that precedent.

jeremydmiller:
Closes #673.

Adds `IDocumentSessionOperations.PendingStreams` — the `StreamAction`s a session has queued but not yet committed.

## The gap

All three stores already surface the *same* `JasperFx.Events.StreamAction` collection; only the spelling of the accessor differs:

| store | spelling |
|---|---|
| Marten | `PendingChanges.Streams()` |
| Polecat | `PendingChanges.Streams` |
| Fisher | `Events.PendingStreams` (no `PendingChanges` facade) |

So each store satisfies the new member by forwarding to what it already has.

## Decisions worth reviewing

**Committable tier**, for the same reason `Events` is (#669): a stream action can only be pending in a session that can append, and `IDocumentWriteOperations` — the tier a projection's `RaiseSideEffects` receives — cannot append or commit.

**Only the streams**, deliberately not a `PendingChanges` / unit-of-work facade. The three products' change sets diverge well past this one collection, and pending *document* operations are not exposed. The measured need is reading what the session is about to append.

**Throwing default, not an empty one**, per the ⚠️ constraint on the issue. An empty default is indistinguishable from a session with nothing pending, so a store that had not implemented the member would silently drop whatever the consumer derives from these actions. Matches the `LoadAsync<T>(object)` (#665) and `Events` (#669) precedent.

The accessor is for code that did *not* do the appending — `StartStream` and `Append` already return the `StreamAction` at the call site. Every compliance fact therefore reads the collection through a contract-typed session rather than a return value.

## Compliance

`PendingStreamActionsCompliance` is what holds stores to the behavior once the compiler stops. Opt-in, and it reuses `DocumentSessionEventsCompliance`'s event types and stream marker, because there are no pending stream actions without an event store — a document-only implementer does not enroll.

The load-bearing fact is the first one: an empty collection on a fresh session, which a store still on the throwing default cannot produce. The one assertion on `StreamActionType` is split into its own fact so it can be gated alone if a store disagrees about it.

`DocumentSessionOperationsDefaultsTests` covers the default itself, which no real store executes and which the compliance suite never reaches (the in-memory reference store in `EventStoreTests` is document-only).

## Downstream

Marten, Polecat and Fisher each need a one-line forwarding member; none breaks on the bump. Same non-covariance trap as #669 — a session already declaring a pending-streams member of a product-specific shape binds to the default rather than failing to compile.

### DIFF
--- a/src/EventStoreTests/Documents/DocumentComplianceConfigTests.cs
+++ b/src/EventStoreTests/Documents/DocumentComplianceConfigTests.cs
@@ -36,6 +36,19 @@ public class DocumentComplianceConfigTests
         config.StreamIdentity.ShouldBe(StreamIdentity.AsString);
     }
 
+    /// <remarks>
+    /// The other suite that appends by stream key, and the one most likely to be added without the
+    /// declaration — it was written before jasperfx#672 gave it anywhere to say this.
+    /// </remarks>
+    [Fact]
+    public void the_pending_stream_actions_suite_declares_string_stream_identity()
+    {
+        var config = new DocumentComplianceConfig();
+        ExposedPendingStreamActionsCompliance.TheConfiguration(config);
+
+        config.StreamIdentity.ShouldBe(StreamIdentity.AsString);
+    }
+
     [Fact]
     public void a_document_only_suite_leaves_stream_identity_alone()
     {
@@ -54,6 +67,13 @@ public class DocumentComplianceConfigTests
     {
         public static readonly Action<DocumentComplianceConfig> TheConfiguration =
             new ExposedDocumentSessionEventsCompliance().Configuration;
+    }
+
+    private class ExposedPendingStreamActionsCompliance
+        : PendingStreamActionsCompliance<InMemoryDocumentComplianceFixture>
+    {
+        public static readonly Action<DocumentComplianceConfig> TheConfiguration =
+            new ExposedPendingStreamActionsCompliance().Configuration;
     }
 
     private class ExposedDocumentSessionCompliance
--- a/src/EventStoreTests/Documents/DocumentSessionOperationsDefaultsTests.cs
+++ b/src/EventStoreTests/Documents/DocumentSessionOperationsDefaultsTests.cs
@@ -1,0 +1,62 @@
+using System.Linq.Expressions;
+using JasperFx.Events;
+using JasperFx.Events.Documents;
+using Shouldly;
+
+namespace EventStoreTests.Documents;
+
+/// <summary>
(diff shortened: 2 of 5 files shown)


jeremydmiller:
⚠️ Cross-reference: #677 (jasperfx#672) adds `DocumentComplianceConfig.StreamIdentity` and has `DocumentSessionEventsCompliance` declare `AsString`, because appending by stream key on a store defaulting to Guid identity fails.

`PendingStreamActionsCompliance` in this PR appends by string stream key too and reuses that suite's event types, so it needs the same line:

```csharp
config.StreamIdentity = StreamIdentity.AsString;
```

The two branches are independent and both cut from `main`. Whichever merges second needs that added — say the word and I'll push it to the trailing branch.

PR #702, opened 2026-08-26
Don't pause a tenant agent forever over a high-water seed below its own position

erdtsieck:
**The problem.** A tenant-scoped agent that is seeded with a high-water mark below its own committed position dies, permanently. `SubscriptionAgent.Apply`'s `Start` branch throws, which funnels into `ReportCriticalFailureAsync` → `AgentStatus.Paused` → and under a node-distributed daemon nothing ever restarts a paused agent (Wolverine's `NodeAgentController.StartAgentAsync` returns early for anything that is not `Stopped`, and `EventSubscriptionAgent.CheckHealthAsync` returns `Unhealthy` before it ever reaches its stall detector / auto-restart). One momentarily missing ceiling therefore takes a tenant's projection out until the next deployment.

**The proposal.** Two small changes, one per cause:

1. `SubscriptionAgent.Apply` **clamps** such a seed to the committed position and logs a warning, instead of throwing. Nothing is replayed or skipped — loading only ever starts above `LastCommitted`, and the mark can only be raised afterwards (`Apply` ignores a lower `MarkHighWater`), so an agent whose seed equals its position simply idles until the coordinator routes it the real mark.
2. `PolledTenantSet` gains a reference-counted `Pin`/`Unpin` that a wholesale `SetTenants` cannot drop, and `StartAgentAsync`'s per-tenant branch pins the tenant across its priming poll **and** its start.

### Why the guard's premise no longer holds

Since marten#4717 the value handed to `Command.Started` for a tenant agent is not the store-wide high water but `_tenantHighWater.CeilingFor(tenantId) ?? 0L` — a per-tenant *snapshot taken at start*. That snapshot can be:

- **missing** (`?? 0L`), because `syncTenantPolling()` rebuilds the polled set from the agents *registered* on the node — which by definition excludes an agent whose start is still in flight. Any concurrent start or stop that reconciles between `Activate(tenant)` and the poll's `PolledTenants.Snapshot()` drops the tenant, and `VectorizedHighWaterMonitor` then silently skips it (`TryGetStatistics` miss → no `_current` entry → `CeilingFor` null);
- **stale**, because a tenant's mark deliberately holds below a sequence gap until `StaleSequenceThreshold` lets the detector skip it, while the projection may already have been walked past that gap by an earlier process.

A committed position above such a snapshot is not evidence of corruption, so it should not be fatal.

For a `SubscribeFromPresent` shard the missing-ceiling case is worse than a pause: `FromPresent` resolves "present" to the mark it is handed, so a seed of 0 makes `ShouldUpdateProgressFirst` rewind the progression row to 0 and replay the tenant's entire history — the regression `primes_the_tenant_ceiling_before_determining_the_starting_position` pins for the sequential path, reintroduced by the race. Hence fixing the race as well as the fallout.

### Field evidence

Production, 2026-08-26, one deployment: 2,173 tenants over 512 shard databases, Wolverine-managed distribution, 3 nodes × ~7,450 agents, `MaxAgentStartParallelism` 25. JasperFx 2.55.0 / Marten 9.29.0 / Wolverine 6.29.2.

| | shard A | shard B |
|---|---|---|
| agent | `claim_lines/all/v21/<tenant>` | `patientstableid/all/<tenant>` |
| failure | `last committed (973739) … high water mark (0)` | `last committed (2010205) … high water mark (2008443)` |
| events for that tenant | 973,739, `max(seq_id)` 973739, 0 archived | 2,010,205, `max(seq_id)` 2010205, 0 archived |
| tenant's persisted `HighWaterMark:<tenant>` row | **973739**, written 207 ms *after* the failing `Start` | 2010205, written 3 min after |

So the data was intact on both sides in both cases; the only thing wrong was the number the agent was seeded with. A: the priming poll produced no reading for that tenant, yet a poll 207 ms later did — the polled-set window. B: the ceiling was 1,762 events behind a progression row that was itself correct.

Two occurrences over ~22,000 agent starts in 24 h, both inside a rollout's start storm. Both agents stayed `Paused` and were still reporting `Unhealthy` every minute hours later; a fleet-wide sweep of `mt_event_progression` found exactly one `Paused` shard out of 22,346 assignments — this one. Neither Wolverine's node-local auto-restart nor its GH-3888 release-to-a-peer can reach a paused agent, and CritterWatch's `restart` cannot either (it only lifts a `Paused` *restriction*, and this shard has none — the daemon paused it).

### Tests

- `tenant_agent_with_no_polled_ceiling_seeds_from_its_own_committed_position` and `tenant_agent_with_a_polled_ceiling_behind_its_progression_still_starts` — the two field cases, red before the change with the exact production message.
- `PolledTenantSetTests` — pinning survives a wholesale `SetTenants`, pins are reference counted (several shards of one tenant start at once), and `Unpin` does not itself remove the tenant.
- `a_tenant_stops_being_polled_once_its_last_agent_stops` — the release half, so a forgotten `Unpin` cannot leave the coordinator polling and persisting rows for a tenant this node runs nothing for.

Full `EventTests` (871), `EventStoreTests` (136) and `CoreTests` (585) pass.

### Alternatives considered

- **Fail the start instead of clamping.** A failed start *is* retried (the leader re-places it) while a paused agent is not, so that would also avoid the permanent stall. It throws away a start for a condition that is harmless, though, and it does not help the `FromPresent` rewind. Happy to switch if you prefer the louder shape.
- **Leave the throw and only fix the race.** The stale-ceiling case (shard B above) is reachable without any race at all, so the guard would keep killing shards.
- The `?? 0L` fallback is deliberately left in place: with the clamp it is no longer fatal, and there is no better value available at that moment.

### DIFF
--- a/src/EventTests/Daemon/HighWater/PolledTenantSetTests.cs
+++ b/src/EventTests/Daemon/HighWater/PolledTenantSetTests.cs
@@ -1,0 +1,69 @@
+using JasperFx.Events.Daemon.HighWater;
+using Shouldly;
+
+namespace EventTests.Daemon.HighWater;
+
+// The polled set is reconciled wholesale from the agents registered on a node, so anything that is
+// merely *starting* has to be held in it explicitly. These pin the two halves of that.
+public class PolledTenantSetTests
+{
+    [Fact]
+    public void set_tenants_keeps_a_pinned_tenant()
+    {
+        var set = new PolledTenantSet();
+
+        // t1's agent start is in flight, so it is not in the assignment snapshot below yet
+        set.Pin("t1");
+        set.SetTenants(["t2"]);
+
+        set.IsPolled("t1").ShouldBeTrue();
+        set.IsPolled("t2").ShouldBeTrue();
+    }
+
+    [Fact]
+    public void pins_are_reference_counted_per_tenant()
+    {
+        var set = new PolledTenantSet();
+
+        // Two shards of the same tenant starting at once
+        set.Pin("t1");
+        set.Pin("t1");
+
+        set.Unpin("t1");
+        set.SetTenants([]);
+
+        // The second start is still running, so the tenant must still be polled for it
+        set.IsPolled("t1").ShouldBeTrue();
+
+        set.Unpin("t1");
+        set.SetTenants([]);
+
+        set.IsPolled("t1").ShouldBeFalse();
+    }
+
+    [Fact]
+    public void unpinning_does_not_itself_remove_the_tenant()
+    {
+        var set = new PolledTenantSet();
+
+        set.Pin("t1");
+        set.Unpin("t1");
+
+        // Whether it belongs in the set is the next reconciliation's call, not the pin's — a start that
+        // succeeded has an agent by now, and dropping it here would stop polling a live tenant.
+        set.IsPolled("t1").ShouldBeTrue();
+    }
+
+    [Fact]
+    public void unpinning_a_tenant_that_was_never_pinned_is_a_no_op()
+    {
+        var set = new PolledTenantSet();
+
+        set.Unpin("t1");
+
+        set.Activate("t2");
+        set.SetTenants(["t2"]);
+
+        set.Snapshot().ShouldBe(["t2"]);
+    }
+}
--- a/src/EventTests/Daemon/PerTenantStartAgentAsyncTests.cs
+++ b/src/EventTests/Daemon/PerTenantStartAgentAsyncTests.cs
@@ -135,6 +135,34 @@ public class PerTenantStartAgentAsyncTests
         agent.LastCommitted.ShouldBe(2010205);
     }
 
+    [Fact]
+    public async Task a_tenant_stops_being_polled_once_its_last_agent_stops()
+    {
+        // The release half of the in-flight start pin: a tenant held in the polled set for the duration
+        // of its start has to be let go again, or the coordinator keeps polling — and persisting
+        // high-water rows for — a tenant this node runs nothing for.
+        var detector = new StubPartitionedDetector();
+        detector.SetTenantMark("t1", 42);
+        detector.SetTenantMark("t2", 42);
+
+        await using var harness = new DaemonHarness(detector,
+            settings => settings.SlowPollingTime = 50.Milliseconds(),
+            shardFor(new ShardName("Trip")));
+
+        await harness.Daemon.StartAgentAsync("Trip:All:t1", CancellationToken.None);
+        await harness.Daemon.StartAgentAsync("Trip:All:t2", CancellationToken.None);
+
+        await harness.Daemon.StopAgentAsync("Trip:All:t1");
+
+        var deadline = DateTimeOffset.UtcNow.AddSeconds(10);
+        while (detector.LastPoll.Contains("t1") && DateTimeOffset.UtcNow < deadline)
+        {
+            await Task.Delay(25, TestContext.Current.CancellationToken);
+        }
+
+        detector.LastPoll.ShouldBe(["t2"]);
+    }
+
     // The Start command is applied on the agent's own command loop, so both the seeding and any failure
     // land just after StartAgentAsync returns. Wait for whichever arrives first rather than sleeping.
     private static async Task<SubscriptionAgent> startedAgentAsync(DaemonHarness harness)
@@ -603,6 +631,18 @@ public class PerTenantStartAgentAsyncTests
             lock (_tenantPolls)
             {
                 return _tenantPolls.SelectMany(x => x).Distinct().ToList();
+            }
+        }
+
+        // The tenant ids the most recent vectorized poll was asked about
+        public string[] LastPoll
+        {
+            get
+            {
+                lock (_tenantPolls)
+                {
+                    return _tenantPolls.Count == 0 ? [] : _tenantPolls[^1];
+                }
             }
         }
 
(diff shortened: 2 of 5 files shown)


jeremydmiller (commented):
**Verdict: correct, and I'd merge it.** Both changes hold up against the code, and the field diagnosis is accurate. I checked out the branch and verified the two new tests are genuine regressions rather than tests written around the fix — on the tests-only commit (2079bf0) they fail with the exact production string, `The last committed number (973739) cannot be higher than the high water mark (0)`. Full branch is 871/871 `EventTests`, three consecutive runs.

## What I confirmed rather than took on faith

**The clamp neither replays nor skips.** After it, `HighWaterMark == LastCommitted == LastEnqueued` and `_bufferedCeiling == LastCommitted`, so `Apply` returns at `if (LastCommitted >= ceiling)` before any load is issued. A later `MarkHighWater` advances normally and `loadNextAsync` picks up above `LastEnqueued`, so nothing is missed either. The optimized-replay branch further down that same `case` can't interact — it requires `LastCommitted == 0`, which a clamp can never coexist with, since the clamp only fires when `LastCommitted >= 1`.

**The pin window is exactly right, not approximately.** `tryStartAgentAsync` registers the agent into `_agents` *and* calls `syncTenantPolling()` inside `_semaphore` before returning, so by the time `startContinuousShardAsync` returns and the `finally { Unpin }` runs, the tenant is already in the assignment snapshot on its own merit. The failure path is covered by `if (!tenantStarted) syncTenantPolling()`, and the `finally` also survives an `OperationCanceledException` out of the semaphore wait at shutdown. I tried to construct a pin leak and couldn't.

**No coalescing hazard.** The start path calls `pollTenantHighWaterAsync()` → `PollAndRouteAsync` (the non-coalesced entry), so each start gets a poll whose `PolledTenants.Snapshot()` is taken *after* its own pin. Routing this through `PollAndRouteCoalescedAsync` would have quietly reintroduced the race by attaching a start to an in-flight poll whose snapshot predates the pin — worth a comment on that line so nobody "tidies" it later.

**The fix is durable, not momentary.** `VectorizedHighWaterMonitor._current` is never pruned — entries only advance monotonically — so once a tenant is carried through one poll by the pin, `CeilingFor` can never return null for it again in that process. That also confirms the shard-A diagnosis: the tenant had never appeared in *any* completed poll's snapshot on that node, not merely a recent one.

## One finding that strengthens the case: this is not a tenancy-only bug

`StartAgentAsync` primes the store-global detector with:

```csharp
if (!_highWater.IsRunning)
{
    await StartHighWaterDetectionAsync().ConfigureAwait(false);
}
```

and `HighWaterAgent.StartAsync` sets `IsRunning = true` **before** awaiting `Detect()`, outside any lock. So under concurrent starts, a second `StartAgentAsync` sees `IsRunning == true`, skips the priming, reaches `tryStartAgentAsync`, and reads `Tracker.HighWaterMark` — still `0`, because the tracker is only written when `Detect()` completes and publishes. `SubscriptionAgent.StartAsync` then posts `Command.Started(request.StartingHighWater ?? _tracker.HighWaterMark, request.Floor)`, and `StartingHighWater` is only set for tenant agents. Meanwhile the default `CatchUp` strategy returns `ProjectionProgressFor(name)` **independent of the `highWaterMark` argument**, so `Floor` is the real persisted position.

Same throw, same `ReportCriticalFailureAsync` → `Paused`, same permanent stall — on a **plain non-tenanted store** under Wolverine's parallel agent starts. Which means:

1. the clamp should stay **global** rather than be narrowed to tenant-bearing agents, and
2. the blast radius of the bug being fixed here is wider than "sharded tenancy at 2,173 tenants" — this is just the deployment large enough to hit it twice in a day and notice.

Might be worth a line in the PR description, since it changes how urgent this looks.

## Three smaller things

- **`PolledTenantSet.Deactivate` still removes a pinned tenant.** It has no production callers today (only `VectorizedHighWaterTests`), so this is latent rather than live — but it is public API on the class whose entire job is now "the set that survives reconciliation", and the next caller who reaches for the obvious-looking method reintroduces exactly the field failure. Cheapest fix is to make it a no-op returning `false` while pinned, or to internalize it given nothing calls it.

- **The same window exists one method over.** `catchUpPerTenantAsync` does the same `Activate` → poll → `CeilingFor(tenantId) ?? 0L` and then `if (ceiling == 0L) continue;` — silently skipping that tenant's catch-up rather than pausing. The per-tenant rebuild path falls back to `Tracker.HighWaterMark` instead. Both are lower severity than the fatal case (the non-pruning cache usually covers them by then, and `ceiling == 0` is also a legitimate "no events yet"), but now that `Pin` exists both are two-line changes. Fine as a follow-up issue rather than in here.

- **The guard was the only detector for "progression row ahead of the event store".** Clamping is the right call — the outcome was wildly disproportionate to the cause — but the cost is worth naming: a genuinely corrupt or over-advanced progression row now produces one `LogWarning` and nothing else, and `Metrics.UpdateGap(HighWaterMark, LastCommitted)` reports a gap of **0** immediately after a clamp, so the agent reads as perfectly caught up on every dashboard. Given the whole point of this PR is that the last one took hours of forensics to find, a counter or a `ShardState` alongside the warning would keep the genuinely-bad case findable.

## On the alternative you offered

I'd keep the clamp over failing the start. A failed start does get retried by the leader, but retry doesn't actually help shard B: the safe-zone hold below the sequence gap is deterministic on restart, so the leader would churn starts until `StaleSequenceThreshold` releases rather than recover. The clamp is harmless in both cases and recovers immediately in both.

## Nit

`unpinning_a_tenant_that_was_never_pinned_is_a_no_op` doesn't assert much — `t1` was never in `_tenants` to begin with, so the `Snapshot()` would be `["t2"]` whether `Unpin` were a no-op or not. It'd bite harder as `Pin("t1")` / `Unpin("t1")` / `Unpin("t1")` / `SetTenants([])`, which pins that the count can't go negative and resurrect the tenant on a later `Pin`/`Unpin` pair.

---

*Reviewed with [Claude Code](https://claude.com/claude-code) — branch checked out, tests run at both commits.*

Issue #503, opened 2026-07-10
Tenant-scoped overloads for the IEventStore explorer read surface (ReadStreamAsync / GetRecentStreamsAsync / GetStreamMetadataAsync)


jeremydmiller:
From CritterWatch#371 Track B1 (event-store explorer tenant-scoping).

## Problem

The `JasperFx.Events.IEventStore` explorer surface is tenant-blind:

```csharp
IAsyncEnumerable<EventRecord> ReadStreamAsync(string streamId, CancellationToken ct)
Task<IReadOnlyList<StreamSummary>> GetRecentStreamsAsync(int count, CancellationToken ct)
Task<StreamMetadata?> GetStreamMetadataAsync(string streamId, CancellationToken ct)
```

On a conjoined multi-tenant store (`TenancyStyle.Conjoined`, with or without `UseTenantPartitionedEvents`), the same stream id can exist under two tenants. A monitoring tool reading through the abstraction gets whichever tenant the store's default session resolves — an ambiguous, potentially **cross-tenant read**. There is also no way to scope "recent streams" to one tenant, and no per-tenant event-store head sequence for correct explorer lag on a partitioned store.

CritterWatch consumes these strictly through the abstraction (its client package must stay store-agnostic across Marten and Polecat), so it cannot work around this by opening a tenant-scoped Marten session.

## Ask

Tenant-scoped overloads with default implementations that delegate to the existing tenant-less members (so non-tenant stores and older implementations keep working unchanged), e.g.:

```csharp
IAsyncEnumerable<EventRecord> ReadStreamAsync(string streamId, string? tenantId, CancellationToken ct)
    => ReadStreamAsync(streamId, ct);
Task<IReadOnlyList<StreamSummary>> GetRecentStreamsAsync(int count, string? tenantId, CancellationToken ct)
    => GetRecentStreamsAsync(count, ct);
Task<StreamMetadata?> GetStreamMetadataAsync(string streamId, string? tenantId, CancellationToken ct)
    => GetStreamMetadataAsync(streamId, ct);
```

plus Marten/Polecat overrides that open the tenant-scoped session when `tenantId` is non-null. A per-tenant head-sequence surface (for explorer lag on `UseTenantPartitionedEvents` stores) would round it out — possibly as `FetchMaxEventSequenceAsync(tenantId)` or as part of the projection-status shape.

[...]

PR #729, opened 2026-09-01
gh-728: a projection-run CLI command

### LINKED ISSUE #728: JasperFx.Events: a projection-run CLI command (step a projection over a stream slice, per-event before/after state)
Companion to JasperFx/CritterWatch#1184 (exposing CritterWatch's projection stepper over MCP). Proposal per Jeremy: the CLI half can live directly in JasperFx.Events rather than in CritterWatch's `cw-*` command set.

**Why here:** CritterWatch's `RequestProjectionRunHandler` is backed entirely by the JasperFx.Events seam — `IEventStore.ReadStreamAsync`, `QueryByTagsAsync`, and `RunProjectionByNameAsync(projectionName, identity, events, startingState, ct)`. Nothing about the replay itself needs CritterWatch; a JasperFx.CommandLine command here would give **every** Marten/Polecat/Fisher application a projection debugger with no console installed, and CritterWatch's MCP tool becomes the fleet-scoped, RBAC-gated remote flavour of the same capability.

**Sketch:**
```
dotnet run -- projection-run <ProjectionName> --stream <id> [--from <v> --to <v>] [--tags k=v ...] [--tenant <id>] [--store <uri>] --json
```
- Source-slice discrimination mirroring `ProjectionRunSourceMode` (Stream / StreamSlice / TagQuery) with the same per-mode validation rules.
- Output: one timeline per resulting aggregate identity, each step carrying event type, timestamp, and before/after state JSON — the shape agents and scripts consume; `--json` by default for agents, a human table otherwise.
- Projection referenced by registered name so the command never needs the consumer's CLR projection types beyond what the host already loads.

**Consumers lined up:** the `critterstack-projection-troubleshooting` AI skill (JasperFx/ai-skills) will teach both this command (dev inner loop, embedded/Fisher hosts) and the CritterWatch MCP tool (production fleets) as one workflow. An intermediate ai-skills release is planned around the projection debugger story.

Open questions: whether `RunProjectionByNameAsync`'s single-identity timeline should grow the multi-bucket surface here or stay 1-bucket until upstream grows it; output-size caps for large slices.

jeremydmiller:
Closes #728.

`dotnet run -- projection-run <Projection> --stream <id>` replays one projection over a source slice and prints the per-event before/after state. Stateless — nothing is written, and the projection's own stored state is never touched.

The capability already existed on `IEventStore`; what was missing was the local half. JasperFx.CommandLine discovery gives the command to every host referencing JasperFx.Events on a binding roll-forward — no Marten/Polecat/Fisher release involved.

### Source modes

```
projection-run Trips --stream trip-1
projection-run Trips --stream trip-1 --from 3 --to 7
projection-run Enrollments --tag:course c-1 --tag:student s-9
```

Plus `--tenant`, `--store`, `--max-events` (default 1000), and `--json`.

The three modes and their per-mode validation rules are deliberately identical to CritterWatch's `RequestProjectionRunHandler`, so an operator who learns one has learned the other.

### Answering the issue's open questions

- **Multi-bucket vs single-identity.** The issue's caveat is stale: upstream already grew the multi-aggregate surface. This sits on `RunMultiStreamProjectionAsync`, so a multi-stream projection reports one timeline per identity and a single-stream projection reports exactly one. That is also what CritterWatch's handler already calls.
- **Output-size caps.** `--max-events` caps the source read at 1000, and the report carries `truncated` — a capped replay otherwise produces a timeline that looks complete.

### Deliberate deviations from the sketch

- Tags use the parser's native `--tag:name value` dictionary-flag shape rather than a hand-parsed `--tags k=v`.
- Every flag is long-form only. `InputParser` derives one-letter aliases from a flag's first letter and has **no collision detection**, so `--stream`/`--store` and `--to`/`--tenant` would silently bind whichever handler was built first. `ProjectionRunCommandParsingTests` is what would notice a lost `longAliasOnly`.
- `--stream` combined with `--tag` is refused. CritterWatch's handler ignores the stream id in tag mode because its UI never sends both; an operator typing both at a prompt means something the command cannot honor.
- An empty result reports zero aggregates rather than CritterWatch's synthetic single bucket, which exists to keep one tab in the SPA.

### Tenancy

The source read always calls the tenant-scoped overloads: a null tenant delegates to the store-global member by contract (#503/#555), so there is no second code path, and `--tenant` against a single-tenant store is refused rather than silently served from whichever tenant the default session happened to hold. Note the fold itself (`RunMultiStreamProjectionAsync`) carries no tenant parameter upstream, so enrichment/grouping still run against the store's default session.

### Tests

952 EventTests green, 44 new: the validation matrix, store selection (including refusing an ambiguous `--store` rather than guessing), slice bounds and truncation against a fake store, the report shape, and the command line itself parsing every flag.

https://claude.ai/code/session_01EVQ4LcDmK6cmvnY792tUEm

### DIFF
--- a/src/EventTests/CommandLine/ProjectionRunCommandParsingTests.cs
+++ b/src/EventTests/CommandLine/ProjectionRunCommandParsingTests.cs
@@ -1,0 +1,71 @@
+using System.Collections.Generic;
+using JasperFx.CommandLine;
+using JasperFx.Events.CommandLine;
+using Shouldly;
+
+namespace EventTests.CommandLine;
+
+/// <summary>
+/// jasperfx#728 — the command line itself. The parser derives one-letter aliases from a flag's first
+/// letter and has NO collision detection, so <c>--stream</c>/<c>--store</c> and <c>--to</c>/<c>--tenant</c>
+/// would silently bind whichever handler was built first. Every flag is long-form only for that
+/// reason, and these tests are what would notice if one lost its <c>longAliasOnly</c>.
+/// </summary>
+public class ProjectionRunCommandParsingTests
+{
+    private static ProjectionRunInput parse(params string[] args)
+    {
+        var command = new ProjectionRunCommand();
+        return (ProjectionRunInput)command.Usages.BuildInput(new Queue<string>(args), new ActivatorCommandCreator());
+    }
+
+    [Fact]
+    public void the_projection_name_is_the_argument()
+    {
+        parse("Trips", "--stream", "trip-1").ProjectionName.ShouldBe("Trips");
+    }
+
+    [Fact]
+    public void stream_and_store_do_not_collide()
+    {
+        var input = parse("Trips", "--stream", "trip-1", "--store", "marten://main");
+
+        input.StreamFlag.ShouldBe("trip-1");
+        input.StoreFlag.ShouldBe("marten://main");
+    }
+
+    [Fact]
+    public void to_and_tenant_do_not_collide()
+    {
+        var input = parse("Trips", "--stream", "trip-1", "--from", "2", "--to", "5", "--tenant", "acme");
+
+        input.FromFlag.ShouldBe(2);
+        input.ToFlag.ShouldBe(5);
+        input.TenantFlag.ShouldBe("acme");
+        input.SourceMode.ShouldBe(ProjectionRunSourceMode.StreamSlice);
+    }
+
+    [Fact]
+    public void tags_are_repeatable_pairs()
+    {
+        var input = parse("Enrollments", "--tag:course", "c-1", "--tag:student", "s-9");
+
+        input.TagFlag["course"].ShouldBe("c-1");
+        input.TagFlag["student"].ShouldBe("s-9");
+        input.SourceMode.ShouldBe(ProjectionRunSourceMode.TagQuery);
+    }
+
+    [Fact]
+    public void json_is_a_boolean_flag()
+    {
+        parse("Trips", "--stream", "trip-1", "--json").JsonFlag.ShouldBeTrue();
+        parse("Trips", "--stream", "trip-1").JsonFlag.ShouldBeFalse();
+    }
+
+    [Fact]
+    public void the_event_cap_defaults_to_a_thousand()
+    {
+        parse("Trips", "--stream", "trip-1").MaxEventsFlag.ShouldBe(1000);
+        parse("Trips", "--stream", "trip-1", "--max-events", "25").MaxEventsFlag.ShouldBe(25);
+    }
+}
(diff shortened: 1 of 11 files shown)


jeremydmiller:
Pushed `954ead3` for the schema reservation — good catch, and it was a hole rather than a polish item.

**The mechanism.** `projection-run` builds the host but never *starts* it, so nothing that normally migrates on startup has run. Whether that matters is per-store and therefore invisible:

| store | on-demand? |
|---|---|
| Marten | document **and** event tables, under the default `AutoCreate.CreateOrUpdate` |
| Polecat | document tables yes; **event tables come from the resource model** |
| Fisher | on demand |
| anything on `AutoCreate.None` | nothing at all |

Same command, works on one developer's box, fails on the next with a raw `relation "mt_events" does not exist` that says nothing about the fix.

**The fix, store-agnostic and needing no new abstraction:** on failure, enumerate `ISystemPart` out of the container, `FindResources()`, and `Check()` each one — throwing *is* how a stateful resource reports "not configured". Failing resources are named and both remedies printed:

```
Unable to read the source events: 42P01: relation "mt_events" does not exist

This application's storage is not ready — 'resources check' fails for: Marten 'Event Store'.
That is the usual reason a projection cannot be replayed: projection-run builds the host but
never starts it, so nothing that normally migrates on startup has run.

projection-run does not migrate anything itself, deliberately: a read-only diagnostic that
quietly changes a database is worse than one that fails. Apply the schema first, then re-run.

  dotnet run -- resources setup   (every registered resource; works on every store)
  dotnet run -- db-apply          (Weasel schema migration; Marten, and Polecat 5.20+)
```

`resources setup` is listed **first** because it works on every store, where `db-apply` only sees a Polecat store from **5.20.0** (polecat#501). That ordering is the one thing here that would be wrong if it were alphabetical.

**Deliberately not auto-migrating.** Someone will eventually point this at production. The message says why, so the next person doesn't add it.

Three properties worth keeping:

- **Failure path only.** Up front it would cost a round trip on every successful run, and would report a broker that happens to be down as though it were why the projection couldn't replay.
- **Never throws.** The caller already has a real error; a failure to *explain* it must not replace it. A diagnosis that fails is simply absent.
- **Bounded** (10s, adjustable). It's a network call diagnosing a failure that already happened.

`ProjectionRunReport` gains a `Diagnosis` field rather than folding the remedy into `Error`, so a `--json` consumer can act on it without parsing prose.

**958 EventTests green, 6 new.**

The companion half is the ai-skills#127 side — teaching an agent to recognise this failure and reach for these commands. That skill is authored against the *published* command, so it follows the release rather than this PR; I've noted your point on it so it isn't left to the raw exception text.

PR #772, opened 2026-09-06
Natural key uniqueness: a second claimant is refused (jasperfx#764 ruling)

jeremydmiller:
Settles the one open product question left behind by jasperfx#764 / #765, and pins it.

## The ruling

**A natural key already mapped to a live stream is refused to a second claimant, not repointed at it.** Fisher refused with `DuplicateNaturalKeyException`; Polecat's `MERGE`-based lookup write repointed the key at the newcomer, which leaves the original stream in place but unreachable by the identifier it was created with, and reports nothing. A natural key exists to name one stream, so a second claimant is a bug in the caller's key derivation rather than an instruction — and of the two failure modes, the silent one is worse.

Fisher's behaviour is the contract. Polecat carries the behavioural change in **JasperFx/polecat#549**, filed against `NaturalKeyUpsertOperation.cs:72` (and `:56` for the conjoined branch), where the `WHEN MATCHED THEN UPDATE SET {streamColumn} = source.{streamColumn}` arm does the repointing.

## The shared exception

`DuplicateNaturalKeyException` lifted from `Fisher.Exceptions` into `JasperFx.Events`, following the jasperfx#751 pattern exactly:

- **Fisher's message adopted verbatim** as the canonical wording. Fisher is the only store that throws it today, so there is no drift to reconcile — which also means Fisher's adoption is a type-forward rather than a subclass.
- **Public constructors** replacing Fisher's `internal` one, plus the **protected message-overriding ctor** the pattern calls for. Fisher's tests do not pin the message (they assert `Key` and `AggregateType`), but the ctor is there for the same reason the other six have one.
- **Superset properties**: `AggregateType` and `Key` are Fisher's; nullable `ExistingStreamId` / `ClaimingStreamId` are added for a throw site that probes the lookup *before* writing and therefore has both ids in hand — which Polecat's fix will, since it has to detect the existing mapping rather than infer the conflict from a write that affected no rows. Nullable in the same way `ExistingStreamIdCollisionException.AggregateType` is nullable: the store that does not know does not have to.

### Why the root namespace and not `JasperFx.Events.Aggregation`

Worth checking, because the issue text points at `Aggregation` — but only the `[NaturalKey]` / `[NaturalKeySource]` attributes and `NaturalKeyAggregateAttribute` live there. **`NaturalKeyDefinition` and `NaturalKeyBuilder` are already in the root `JasperFx.Events` namespace** (`src/JasperFx.Events/NaturalKeyDefinition.cs`, `NaturalKeyBuilder.cs`), and so are all nine shared event exceptions. The root is where both neighbourhoods already are, so `JasperFx.Events.DuplicateNaturalKeyException` sits beside the type that produces the key value it carries *and* beside the exceptions it is a sibling of. Putting it in `Aggregation` would separate it from both.

## The compliance fact

`NaturalKeyCompliance.a_second_stream_cannot_claim_a_live_natural_key`, under the existing `SupportsNaturalKeys` gate. It asserts two things, and **the second is the one that does the work**:

1. The second stream's `SaveChangesAsync` throws `DuplicateNaturalKeyException`, carrying the unwrapped key and the aggregate type.
2. The **original stream's mapping still resolves afterwards** — read back through `FetchLatest` and compared **by stream id**, not by nullness.

Asserting only the throw would pass on a store that repoints the row and *then* fails the transaction, or that writes the new mapping through a path the failure does not roll back. The mapping is the thing the refusal exists to protect, so it has to be read back; and comparing by id rather than by "something resolves" is what keeps a repointing store from passing.

The fact is deliberately distinct from its two neighbours, both already pinned: re-asserting the same mapping on its own stream is idempotent by design, and `renaming_the_natural_key_retires_the_previous_key` is what legitimately frees an identifier for a later stream to claim.

Also removes the "stays product-owned pending a decision" note from the suite's class remarks and from the README, replacing both with the ruling; the README's suite-table row now names uniqueness.

## Validation

GitHub Actions is stalled org-wide, so this was validated locally:

- `JasperFx.Events.ComplianceTests` builds clean (0 errors, net9.0 + net10.0), before and after merging `origin/main`.
- `EventStoreTests` — 141 passed, 0 failed.
- New `EventTests/DuplicateNaturalKeyExceptionTests` — 3 passed (canonical message and shape, the both-ids ctor, and a subclass keeping a diverged message).

**Honest scope of that**: `EventStoreTests` enrols only the *document* compliance suites — it has no `SupportsNaturalKeys` fixture at all — so it validates compilation and no-regression, **not** the new fact's runtime behaviour. No store is enrolled in `NaturalKeyCompliance`, so the fact cannot go red on merge either; it is the bar Polecat's and Fisher's eventual enrolment will have to clear.

Closes the follow-up left open by jasperfx#764.

### DIFF
--- a/src/EventTests/DuplicateNaturalKeyExceptionTests.cs
+++ b/src/EventTests/DuplicateNaturalKeyExceptionTests.cs
@@ -1,0 +1,66 @@
+using JasperFx.Events;
+using Shouldly;
+
+namespace EventTests;
+
+// jasperfx#764: the natural key uniqueness ruling. Fisher's DuplicateNaturalKeyException is lifted
+// beside the jasperfx#751 exceptions with its message adopted verbatim, so Fisher's swap is a
+// type-forward rather than a subclass. The tests below pin the canonical shape a store has to keep
+// expressible.
+public class DuplicateNaturalKeyExceptionTests
+{
+    [Fact]
+    public void carries_the_aggregate_type_and_the_key()
+    {
+        var ex = new DuplicateNaturalKeyException(typeof(Order), "ORD-1234");
+
+        ex.AggregateType.ShouldBe(typeof(Order));
+        ex.Key.ShouldBe("ORD-1234");
+
+        ex.Message.ShouldBe(
+            "The natural key 'ORD-1234' is already mapped to a different stream for aggregate type "
+            + "'Order'. A natural key identifies one stream; if the mapping is meant to move, delete "
+            + "the existing row first.");
+
+        // Fisher's throw site infers the conflict from a write that affected no rows, so it has
+        // neither stream id in hand. Both stay optional.
+        ex.ExistingStreamId.ShouldBeNull();
+        ex.ClaimingStreamId.ShouldBeNull();
+    }
+
+    [Fact]
+    public void carries_both_stream_ids_when_the_throw_site_knew_them()
+    {
+        // The shape a store takes when it probes the lookup before writing rather than reacting to
+        // an empty result — it knows who holds the key and who tried to take it.
+        var existing = Guid.NewGuid();
+        var claiming = Guid.NewGuid();
+
+        var ex = new DuplicateNaturalKeyException(typeof(Order), "ORD-1234", existing, claiming);
+
+        ex.ExistingStreamId.ShouldBe(existing);
+        ex.ClaimingStreamId.ShouldBe(claiming);
+        ex.Key.ShouldBe("ORD-1234");
+    }
+
+    [Fact]
+    public void a_store_subclass_can_keep_its_diverged_message()
+    {
+        var ex = new StoreishDuplicateNaturalKeyException(typeof(Order), "ORD-1234");
+
+        ex.Message.ShouldBe("Natural key ORD-1234 is taken");
+        ex.AggregateType.ShouldBe(typeof(Order));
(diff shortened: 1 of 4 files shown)


jeremydmiller:
Two notes on the record, post-merge.

**A correction to the description.** It says Fisher's adoption is "a type-forward rather than a subclass" — that is loose. Type forwarding requires the same fully-qualified name, and the namespaces differ (`Fisher.Exceptions.DuplicateNaturalKeyException` vs `JasperFx.Events.DuplicateNaturalKeyException`), so `[assembly: TypeForwardedTo]` is not actually available. What is true, and what the point was: because the canonical message and shape were taken verbatim from Fisher's copy, there is no drift to reconcile — Fisher either deletes the local type and throws the shared one, or keeps it as a plain subclass whose ctor calls the two-arg base rather than the message-overriding one. The protected message-overriding ctor is there for the pattern's sake, not because Fisher needs it.

**Fisher follow-up filed as JasperFx/fisher#178**, laying out those two options (the subclass keeps existing `catch (Fisher.Exceptions.DuplicateNaturalKeyException)` clauses working; deleting is tidier but source-breaking) and noting that the local `UnknownNaturalKeyException` in the same file stays put, since the `FetchForWriting` miss is the one thing all three stores genuinely spell differently.

PR #782, opened 2026-09-06
Fix three wave-13 compliance facts that cannot pass on any store

### LINKED ISSUE #781: Two wave-13 compliance facts cannot pass on any store (wrong projection name; contradictory initial-state assertions)
Two facts in the wave-13 compliance suites cannot pass on **any** store. Both were found by the first execution of these suites against a real event store — Polecat's adoption in [polecat#556](https://github.com/JasperFx/polecat/issues/556). Neither is store-specific; both would fail identically on Marten and Fisher.

## 1. `ProjectionSideEffectCompliance` rebuilds by the wrong projection name

`side_effects_are_suppressed_during_a_rebuild` and `no_messages_are_published_during_a_rebuild` both call:

```csharp
await daemon.RebuildProjectionAsync(nameof(ComplianceWatchtowerProjection), _timeout, CancellationToken.None);
```

`ProjectionBase`'s constructor defaults `Name` to `GetType().Name`, but `JasperFxAggregationProjectionBase`'s constructor **overrides** it:

```csharp
Name = typeof(TDoc).NameInCode();
```

So an aggregation projection is known to the daemon by the type it *produces*, not by its class. The observed failure is:

```
ArgumentOutOfRangeException : No registered projection matches the name
'ComplianceWatchtowerProjection'. Available names are 'ComplianceWatchtower'
```

That is shared code, so every consumer sees it. Fix: name the document type.

## 2. `AggregateToLinqOperatorCompliance.can_aggregate_with_initial_state_asynchronously` asserts a self-contradictory outcome

```csharp
var initial = new ComplianceTrail { Miles = 100 };
var trail = await theFixture.AggregateEventsToAsync(session, e => e.StreamId == stream, initial, Cancellation);

trail.Name.ShouldBe("Long Trail");   // requires Create(TrailStarted) to have run
trail.Miles.ShouldBe(110);           // requires the supplied state to have survived
```

`ComplianceTrail.Create(TrailStarted e) => new() { Name = e.Name }` is a **creator**: it returns a fresh instance. The two assertions are mutually exclusive —

- if the creator runs, the supplied state is discarded and `Miles` is 10, not 110;
- if the supplied state is kept (the correct semantics for "fold onto this state" — the aggregate already exists, so no creator is invoked), `Name` is never assigned and stays `""`.

There is no implementation that yields both. The store under test returned `Name == ""`, `Miles == 110`, which is the coherent answer.

Fix: keep `Miles == 110`, which is what the fact is actually about, and pin `Name == string.Empty` as the documented consequence — a supplied initial state suppresses the creator.

## Note

Neither fix weakens a suite. The first corrects an identifier; the second replaces an unsatisfiable pair of assertions with the behaviour that is actually contractual, and pins it in both directions rather than dropping it.

jeremydmiller:
Closes #781.

Three facts in the wave-13 compliance suites could not pass on **any** store. All surfaced the first time these suites were executed against a real event store — Polecat's adoption in [polecat#556](https://github.com/JasperFx/polecat/issues/556). Until now they were compile-checked and design-reasoned only, because this repo enrols the document suites but no event store.

## 1. `ProjectionSideEffectCompliance` rebuilt by the wrong name

Both rebuild facts named the projection **class**:

```csharp
await daemon.RebuildProjectionAsync(nameof(ComplianceWatchtowerProjection), _timeout, CancellationToken.None);
```

`ProjectionBase` defaults `Name` to `GetType().Name`, but `JasperFxAggregationProjectionBase`'s constructor overrides it with `typeof(TDoc).NameInCode()` — an aggregation projection is known to the daemon by the type it produces. Both facts threw:

```
ArgumentOutOfRangeException : No registered projection matches the name
'ComplianceWatchtowerProjection'. Available names are 'ComplianceWatchtower'
```

Now names the document type, with a comment so the next reader does not "fix" it back.

## 2. `can_aggregate_with_initial_state_asynchronously` asserted an impossible pair

```csharp
trail.Name.ShouldBe("Long Trail");   // only if Create(TrailStarted) ran
trail.Miles.ShouldBe(110);           // only if the supplied state survived
```

`ComplianceTrail.Create` is a creator — it returns a fresh instance. Run it and the supplied state is gone (`Miles` is 10); skip it and `Name` is never assigned. The two cannot both hold.

Folding onto a supplied state means the aggregate already exists, so skipping the creator is the correct semantics, and it is what the store under test did. The fact now keeps `Miles == 110` — the thing it is actually about — and pins `Name == string.Empty` as the documented consequence. That is **stronger** than dropping the assertion: it now fails a store that restarts from the creator as well as one that ignores the initial state.

## 3. `StreamArchivingCompliance`'s async and string facts appended `Archived` alone

A projection's `AllEventTypes` comes from its `Create`/`Apply` handlers, and `ComplianceLedger` declares no `Archived` handler. So a slice carrying **only** an `Archived`:

- fails `AppliesTo` on the inline path and is screened out before `maybeArchiveStream` is reached;
- never survives the daemon's event-type allow list on the async path.

The inline fact immediately above them already appends `LedgerEntryPosted(75)` alongside its `Archived`, and so do **all four** of Marten's local variants (`archiving_events.cs` — `new DEvent(), new Archived("Complete")`, every time). The two ported facts lost that companion event. Restored.

## Not a weakening

None of these relaxes a suite to accommodate an implementation. The first corrects an identifier; the second replaces an unsatisfiable assertion pair with the contractual behaviour, pinned in both directions; the third restores a precondition the facts always needed and that their own sibling and their Marten ancestors carry.

## Validation

Built Polecat's test assembly against this working copy rather than the package, via Polecat's documented `-p:ComplianceSourceDir=...` flow, and ran the full `Polecat.Tests.Compliance` namespace against SQL Server 2025: **507 passed, 0 failed, 3 skipped** (the 3 skips are capability gates a store legitimately declines). All five previously-failing facts pass; no other fact changed state.

Worth noting these fixes did **not** paper over store bugs — the same run also found five genuine Polecat bugs, fixed separately in polecat#556.

### DIFF
--- a/src/JasperFx.Events.ComplianceTests/Suites/AggregateToLinqOperatorCompliance.cs
+++ b/src/JasperFx.Events.ComplianceTests/Suites/AggregateToLinqOperatorCompliance.cs
@@ -96,8 +96,20 @@ public abstract class AggregateToLinqOperatorCompliance<TFixture, TOperations, T
             e => e.StreamId == stream, initial, Cancellation);
 
         trail.ShouldNotBeNull();
-        trail.Name.ShouldBe("Long Trail");
+
+        // The fold CONTINUES the supplied state rather than restarting from it: 100 + 10.
         trail.Miles.ShouldBe(110);
+
+        // And the creator does not run, which is the other half of the same statement. A supplied
+        // initial state means the aggregate already exists, so ComplianceTrail.Create(TrailStarted)
+        // is skipped and Name is never assigned — TrailStarted contributes nothing but its place in
+        // the sequence.
+        //
+        // This originally asserted Name == "Long Trail" alongside Miles == 110, which no coherent
+        // implementation can satisfy: running the creator produces a NEW instance and discards the
+        // supplied state, so Name and Miles == 110 are mutually exclusive. Caught the first time the
+        // suite was executed against a real store (polecat#556).
+        trail.Name.ShouldBe(string.Empty);
     }
 
     [Fact]
--- a/src/JasperFx.Events.ComplianceTests/Suites/ProjectionSideEffectCompliance.cs
+++ b/src/JasperFx.Events.ComplianceTests/Suites/ProjectionSideEffectCompliance.cs
@@ -297,7 +297,12 @@ public abstract class ProjectionSideEffectCompliance<TFixture, TOperations, TQue
 
         await daemon.StopAllAsync();
 
-        await daemon.RebuildProjectionAsync(nameof(ComplianceWatchtowerProjection), _timeout,
+        // nameof(ComplianceWatchtower), the DOCUMENT type, not the projection class:
+        // JasperFxAggregationProjectionBase's constructor sets Name = typeof(TDoc).NameInCode(),
+        // overriding ProjectionBase's class-name default, so an aggregation projection is known to
+        // the daemon by the type it produces. Naming the class here threw
+        // ArgumentOutOfRangeException on every store (found enrolling this suite in polecat#556).
+        await daemon.RebuildProjectionAsync(nameof(ComplianceWatchtower), _timeout,
             CancellationToken.None);
 
         var after = await eventsForAsync(streamId);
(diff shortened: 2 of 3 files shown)


jeremydmiller:
Closing as superseded — every finding here was independently found and fixed on `main` while this was open, by Marten's parallel adoption of the same wave (marten#5343 → #780, #784).

- The `nameof(ComplianceWatchtowerProjection)` rebuild name: fixed in #780, via a `ComplianceWatchtowerProjection.ProjectionName` constant — better than my `nameof(ComplianceWatchtower)`, because it survives the document type being renamed.
- `can_aggregate_with_initial_state_asynchronously`: fixed in #780, by dropping the `Name` assertion. Mine pinned `Name == string.Empty` instead; #780's remark makes the case for not pinning it at all, and that call is already merged.
- The archiving facts' `Archived`-only slices: #784 fixes this the **right way round**, and my change here was actively wrong. I patched the *suite* to append a handled event alongside the marker; #784 fixes the *framework* so an `Archived`-only slice works — `determineEventTypes` now appends `Archived` for single-stream projections, closing the `AppliesTo` screen and the async shard's event filter in one move. Making the suite avoid the gap would have hidden it.

Nothing here needs to land. Refiling nothing; #781 closed alongside.

For the record, the same Polecat run that produced these three also found five genuine Polecat bugs, which are fixed in polecat#556 and are unaffected by this.

Issue #550, opened 2026-07-21
Stamp AssignedNodeNumber on daemon-published ShardState so managed distribution populates running_on_node (marten#5001)


jeremydmiller:
## Summary

Under Balanced, Wolverine-managed subscription distribution with extended-progression tracking on, `mt_event_progression.running_on_node` stays **NULL** even though the projection agents are genuinely assigned to a real node. The sibling telemetry (`heartbeat`, `agent_status`, `last_updated`) writes correctly in the same row on the same path — only `running_on_node` never populates.

Downstream tracking issue: [JasperFx/marten#5001](https://github.com/JasperFx/marten/issues/5001). Repro (green, pins the gap with a `// FLIP-WHEN-FIXED` marker): CritterWatch `RunningOnNodeMultiNodeTests`.

The load-bearing fix is here in JasperFx.Events (the daemon publish surface). A one-line wire-up in Wolverine (which owns the node assignment) is a companion task, described below. **No Marten change is required.**

## Root cause (traced end to end)

`ExtendedProgressionWriter.OnNext` already has the carry logic, and its own comment says the value is expected to be *"stamped by a distribution layer (e.g. Wolverine-managed subscription distribution)"*:

```csharp
// JasperFx.Events/Daemon/ExtendedProgressionWriter.cs
if (value.RunningOnNode == null && value.AssignedNodeNumber != 0)
    value.RunningOnNode = value.AssignedNodeNumber;
```

But **nothing ever stamps `AssignedNodeNumber` on a runtime-published `ShardState`.** A grep of both trees finds `AssignedNodeNumber = …` in exactly one place — Marten's `ShardStateSelector`, which sets it only when reading the `assigned_node` column back from the DB (and nothing writes that column; it defaults `0`). Every `ShardState` a `SubscriptionAgent` publishes carries `AgentStatus`/`LastHeartbeat` but `AssignedNodeNumber == 0`, so the carry block is dead and `running_on_node` persists NULL. The carry logic is correct; its input is never supplied on the runtime publish path.

## Proposed fix — Part 1 (JasperFx.Events): one stamp point on `ShardStateTracker`

This mirrors exactly how `DatabaseIdentifier` is already stamped for every state the daemon publishes:

```csharp
// ShardStateTracker.cs
public int AssignedNodeNumber { get; set; } // 0 = unset

public ValueTask PublishAsync(ShardState state)
{
[...]

PR #797, opened 2026-09-06
gh-796: a single stream projection applies to Compacted<TDoc>, for every store at once

### LINKED ISSUE #796: Compacted<TDoc> is missing from a single stream projection's determineEventTypes(), so a filtered async shard loses the whole pre-compaction history
Follow-up to #784, and the same defect one type over.

#784 established the rule: `Archived` carries no state, so an aggregate has no reason to declare an
`Apply` for it, and leaving it out of `determineEventTypes()` kept it out of the async shard's event
filter as well as out of `AppliesTo`. `determineEventTypes` now appends it for the single stream
scope.

`Compacted<TDoc>` is in exactly the same position and was not included.

## Why it matters more than `Archived` does

A dropped `Archived` loses an archival. A dropped `Compacted<TDoc>` loses **the stream's entire
pre-compaction history**, because `CompactStreamAsync` *deletes* the events it folds and leaves the
marker in their place. The marker is not an extra event beside the history — it is the only thing
left of it, and `Compacted<T>.MaybeFastForward` is what restarts a fold from it.

So an async single-stream shard for a projection with a declared event list, catching up over a
compacted stream from a null snapshot (a rebuild, or a projection added after compaction), folds
only what was appended *after* compaction and reports an aggregate missing everything before it.
No exception, no log line, and an unfiltered shard over the identical rows disagrees with it.

## Who is affected

Every store, because every store composes the loader's filter from `IncludedEventTypes`:

- **Marten** is fine, but only because it patches around the gap itself, in
  `DocumentStore.EventStore.buildEventLoaderFilters`:
  ```csharp
  var allTypes = filterable.IncludedEventTypes.Concat([typeof(Archived)]).ToArray();
  if (filterable is IAggregateProjection aggregateProjection)
  {
      var compactedType = typeof(Compacted<>).MakeGenericType(aggregateProjection.AggregateType);
      allTypes = allTypes.Concat([compactedType]).ToArray();
  }
  ```
- **Polecat** did not, and that is JasperFx/polecat#557 — fixed store-locally in
  `PolecatEventLoader` for now, with a regression test showing a filtered and an unfiltered load
  disagreeing over one compacted stream.
- **Fisher** does not either. `FisherEventLoader`'s constructor builds `_allowedEventTypeNames`
  straight from `filtering.IncludedEventTypes` with nothing added, so it has the same bug today and
  nothing in its suite pins it.

## Suggested fix

Append `typeof(Compacted<TDoc>)` alongside `Archived` in
`JasperFxAggregationProjectionBase.determineEventTypes()`, under the same single-stream scope guard
and the same "an empty set is left empty" caveat #784 called out — an empty `AllEventTypes` means
"applies to everything" and must stay empty.

That puts it where #784 put its sibling, makes it correct for every store at once, and lets Marten
drop half of its local patch and Polecat drop its own.

If it should stay a per-store responsibility instead, that is worth saying explicitly, because two
of the three stores got it wrong independently and the failure is silent in both.

jeremydmiller:
Follow-up to #784, and the same defect one type over. Closes #796.

#784 established that `Archived` carries no state, so an aggregate has no reason to declare an `Apply` for it, and leaving it out of `determineEventTypes()` kept it out of the async shard's event filter as well as out of `AppliesTo`. `Compacted<TDoc>` was in exactly the same position and was not included.

## Why it is the worse of the two

A dropped `Archived` loses an archival. A dropped `Compacted<TDoc>` loses the stream's **entire pre-compaction history**, because `CompactStreamAsync` *deletes* the events it folds and leaves the marker in their place. The marker is not an extra event beside the history — it is the only thing left of it, and `Compacted<T>.MaybeFastForward` is what restarts a fold from it.

## Why here rather than per store

Every store composes its async loader's allow list from `IncludedEventTypes`. Marten patched around the gap locally in `buildEventLoaderFilters`; Polecat (JasperFx/polecat#557) and Fisher (JasperFx/fisher#203) did not, and got it wrong independently. Both can now drop their local patches.

Same single-stream scope guard and the same empty-set caveat as #784: an empty `AllEventTypes` means "applies to everything", which is how a catch-all `Evolve(IEvent)` projection declares itself, and must stay empty.

## Also: the append was duplicating

`AssembleAndAssertValidity` ends with `IncludedEventTypes.Fill(determineEventTypes())`, so the marker this override appends is written back into `IncludedEventTypes` — and the next evaluation concats that list again, past the base's own `Distinct()`, and appends the marker a second time. `AllEventTypes` came back as `[AEvent, Archived, Archived]`.

It does not change what `AppliesTo` answers, but every store composes its loader's allow list from these types, so the duplicate reached the generated SQL as a repeated `IN` member and a repeated parameter slot. The append is now deduplicated.

## Verification

The failure is silent, so a green suite proves nothing on its own. I packed this branch to a local feed twice — once with the fix, once with the production change alone reverted — and ran Fisher against both.

Without the fix, the new compliance fact fails on Fisher exactly as JasperFx/polecat#567 described:

```
Shouldly.ShouldAssertException : meter.Location
    should be
    but was
```

The document **is** written, so a "document is missing" assertion would not have caught it — only `Location` came back empty. With the fix it passes. **JasperFx/fisher#203 falls out of this, verified rather than assumed.**

The new fact deliberately appends one event *after* the compaction. Without that trailing event a broken shard sees an empty slice and writes no document, which the assertion would catch for the wrong reason; with it, the broken shard writes a *plausible* document and only the totals tell them apart.

## The composite question from the issue

`CompositeProjection` inherits `IncludedEventTypes` from `EventFilterable` and nothing rolls its children's lists up into it, so it stays empty and the composite's shard is unfiltered. The bug cannot occur there today. Worth knowing that it depends on the empty-list-means-no-filter path staying as it is.

## Test results

| Suite | Result |
|---|---|
| EventTests | 1053 green, net9.0 and net10.0 |
| EventStoreTests | 141 green, net9.0 and net10.0 |
| JasperFx.Events.SourceGenerator.Tests | 62 green |
| Fisher.Tests (against the packed build) | 1933 green, net9.0 and net10.0 |
| Polecat.Tests `stream_compacting_compliance` (against the packed build) | 12 green |

### DIFF
--- a/src/EventTests/Projections/SingleStreamProjectionTests.cs
+++ b/src/EventTests/Projections/SingleStreamProjectionTests.cs
@@ -26,6 +26,67 @@ public class SingleStreamProjectionTests
             explanation);
     }
 
+    // jasperfx#778 put Archived into a single stream projection's event types; jasperfx#796 is the
+    // same defect one type over. Every store composes its async loader's filter from
+    // IncludedEventTypes, so a Compacted<TDoc> missing here is a filtered shard that never sees the
+    // marker -- and the marker is the only thing left of the pre-compaction history, because
+    // CompactStreamAsync deletes the events it folds. The shard then folds post-compaction events
+    // onto a null snapshot and persists an aggregate missing everything before the compaction, with
+    // nothing thrown and nothing logged.
+    [Fact]
+    public void declared_event_types_carry_archived_and_compacted_over_the_aggregate()
+    {
+        var projection = new ConventionalProjection();
+        projection.AssembleAndAssertValidity();
+
+        projection.AllEventTypes.ShouldContain(typeof(Archived));
+        projection.AllEventTypes.ShouldContain(typeof(Compacted<MyAggregate>));
+    }
+
+    // Compacted<T> is closed over the projection's OWN aggregate. Appending an open or a
+    // wrong-aggregate marker would widen the shard's filter to markers it has nothing to do with,
+    // which is the same objection that keeps Archived out of the multi stream scope.
+    [Fact]
+    public void the_compacted_marker_is_closed_over_this_projections_aggregate_only()
+    {
+        var projection = new ConventionalProjection();
+        projection.AssembleAndAssertValidity();
+
+        projection.AllEventTypes.ShouldNotContain(typeof(Compacted<>));
+        projection.AllEventTypes.ShouldNotContain(typeof(Compacted<MyOtherAggregate>));
+    }
+
+    // AssembleAndAssertValidity ends with IncludedEventTypes.Fill(determineEventTypes()), so the
+    // marker this override appends is written back into IncludedEventTypes -- and the next evaluation
+    // concats that list again, past the base's own Distinct(), and appends the marker a second time.
+    // Before jasperfx#796 this shipped: AllEventTypes came back [AEvent, Archived, Archived]. It does
+    // not change what AppliesTo answers, but every store composes its async loader's allow list
+    // straight from these types, so the duplicate reaches the generated SQL as a repeated IN member
+    // and a repeated parameter slot.
+    [Fact]
+    public void the_appended_markers_are_not_duplicated_by_the_assembly_fill()
+    {
+        var projection = new ConventionalProjection();
+
+        // The Fill happens here -- one call is enough to seed IncludedEventTypes with the markers.
+        projection.AssembleAndAssertValidity();
+
+        projection.AllEventTypes.ShouldBe(projection.AllEventTypes.Distinct().ToArray());
+    }
+
+    // The empty-set guard from jasperfx#778, restated for the second marker. AppliesTo reads an
+    // empty AllEventTypes as "applies to everything" -- how a catch-all Evolve(IEvent) projection
+    // declares itself -- so appending either marker unconditionally turns that "everything" into
+    // "exactly two types" and the projection then sees nothing at all.
+    [Fact]
+    public void an_empty_event_type_set_stays_empty()
+    {
+        var projection = new CatchAllEvolveProjection();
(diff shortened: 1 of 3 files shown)


jeremydmiller:
## Downstream check: Polecat, against a local pack of this branch

**2560 tests, 4 failures**, all in one class — `Polecat.Tests.Daemon.compaction_marker_allow_list_tests`, the tests that came with Polecat's own local patch (JasperFx/polecat#557 / #567). All four die in the same shared setup helper, before exercising anything:

```
compaction_marker_allow_list_tests.cs:218
    projection.IncludedEventTypes.ShouldNotContain(typeof(Compacted<QuestParty>));
```

That guard encodes the class's premise, which its doc comment states outright — *"Nothing upstream in JasperFx adds it for the store"*. This PR makes that deliberately false, which is the entire point of fixing it here rather than per store.

**The behaviour is right.** The one test in that class that does not go through the helper — `an_async_shard_folds_a_compacted_stream_the_way_live_aggregation_does` — passes, and so does `stream_compacting_compliance` 12/12, including the new fact this PR adds. The four failures never reached the loading they were written to check.

Filed as JasperFx/polecat#568: drop the now-redundant local patch and re-premise those tests, once this publishes. Worth noting there that the three loader-path tests should survive rather than be deleted — the #550 push-down cap and the client-side fallback are Polecat's own code and nothing shared pins them.

Marten's `buildEventLoaderFilters` patch becomes redundant the same way. I have not filed anything there, since Marten is another agent's territory right now.

### Summary of downstream verification

| Store | Result |
|---|---|
| Fisher | 1933 green, both TFMs. Fails without the fix — document written, `Location` empty. |
| Polecat | 2556 green; 4 failures, all one stale precondition guard, behaviour verified correct separately. |

PR #833, opened 2026-09-13
gh-832: order step 3's missing-frame insertion by the method's own frame list

### LINKED ISSUE #832: MethodFrameArranger step 3 walks the dependency cache unsorted, so generated statement order moves between runs
Follow-up to #196. That fix sorted the two consumers of `DependencyGatherer.Variables`, which made constructor parameters and field declarations stable. Step 3 of the same method still walks `DependencyGatherer.Dependencies` unsorted, so the statements inside the generated method body still move between runs.

## What happens

`MethodFrameArranger.compileFrames` step 3 collects the frames that create variables a method needs but does not already hold, and inserts them at the front of the frame list:

```csharp
// Step 3, gather any missing frames and
// add to the beginning of the list
dependencies.Dependencies.SelectMany(x => x).Distinct()
    .Where(x => !frames.Contains(x))
    .Each(x => frames.Insert(0, x));
```

`DependencyGatherer.Dependencies` is a `LightweightCache<Frame, List<Frame>>`, so an `ImHashMap` keyed by `Frame`. `Frame` does not override `GetHashCode`, so the key hash is the identity hash - an arbitrary value per object - and `ImHashMap` enumerates in ascending hash order. Two arrangements of the same logical method visit those dependency lists in different orders and insert the missing frames in different orders. Step 4's topological sort preserves the relative order of frames that do not depend on each other, so nothing downstream restores a stable order.

## What it looks like

Regenerating an application's committed generated code produces diffs that carry no semantic change - the same statements, emitted in a different order. In a representative diff a service location statement is emitted near the top of a generated method in one run and seventeen lines lower in the next, everything else identical. Anything that byte-compares generated output - a drift check in CI, code review, a reproducible build - cannot distinguish that from a real change.

## Provenance

Measured by running `codegen write` twice in the same environment on an application with committed generated code, and comparing the two outputs to each other (not against the committed files - a different configuration shape changes what gets generated, so that comparison would not be valid):

- JasperFx 1.24.1: 15 of 60 generated files differed between the two runs.
- JasperFx 1.31.0, the first 1.x release carrying #196: 6 of 60 still differed.

That version bump is what points at step 3 as what is left: everything #196 covers is stable across those two runs, and what is left moving is statement order.

JasperFx/wolverine#3059 reported the same drift as a macOS-vs-Linux difference and guessed at iteration order of an intermediate frame collection as one of three possible causes, without a repro. This confirms it.

## Repro

Two failing tests written against `CodegenTests`, sitting on a branch on my fork rather than in a PR. No application needed - several independent service location calls feeding two downstream calls, with nothing in the dependency graph ordering the locators relative to each other.

- One generates the same type 26 times in a single process and requires every output to be byte-identical.
- One pins two `Frame` hash codes and swaps them, so it fails deterministically rather than depending on identity hash values.

Both fail on `main` at 2.69.1, on net9.0 and net10.0. I ran them twelve times and they failed every time; the hash-pinned one fails by construction rather than by luck. Nothing else in `CodegenTests` moves - 466 passing before, 468 after.

## Suggested fix

The smallest fix I can see is one statement: walk `frames` and index the cache, instead of enumerating the cache. The keys of that cache are exactly the frames passed in, so the set gathered is unchanged and only the order becomes deterministic.

Both commits are on that branch: https://github.com/AllainPL/jasperfx/tree/fix/codegen-frame-ordering-determinism

I will open it as a PR. Just say if you want it done differently.

AllainPL:
Fixes #832: `MethodFrameArranger.compileFrames` step 3 enumerates `DependencyGatherer.Dependencies`, an `ImHashMap` keyed by `Frame`. `Frame` does not override `GetHashCode`, so that walk follows identity hash codes and differs between two arrangements of the same logical method. Step 4's topological sort preserves the relative order of independent frames, so the emitted body carries the variation. #196 fixed the same thing in the two calls immediately above, for `DependencyGatherer.Variables`, which is why constructor parameters and fields are stable today and the statements inside the method are not.

Commit 1 is the repro, commit 2 is the fix. Happy to rework it if you want it done differently.

## The change

Walk `frames` to collect the missing dependency frames instead of enumerating the cache:

```csharp
var missing = frames.SelectMany(x => dependencies.Dependencies[x]).Distinct()
    .Where(x => !frames.Contains(x))
    .ToArray();

foreach (var frame in missing) frames.Insert(0, frame);
```

The cache's keys are exactly the frames passed in - `DependencyGatherer`'s constructor fills one entry per frame and nothing adds a key before step 3 - so the set gathered is unchanged and only the order becomes deterministic. `ToArray()` is load-bearing now that the enumeration source is `frames` itself; the old lazy form was safe only because it enumerated the cache. Sorting was the other option, but `Frame` has no stable key to sort on, and `OrderBy` is a stable sort, so ties would fall straight back to hash order.

## The tests

`CodeGeneration_created_variable_ordering_determinism`, a companion to `CodeGeneration_ordering_determinism` from #196. Every test in that file builds `InjectedField`s, so the suite covers the constructor and the field declarations but nothing that creates variables. The new ones model that shape with existing primitives: independent service location calls feeding two downstream calls, with nothing in the dependency graph ordering the locators relative to each other.

One generates the same type 26 times in a single process and requires byte-identical output. The other pins two `Frame` hash codes and swaps them, which makes the failure deterministic rather than a coin flip.

Both failed on `main` twelve times out of twelve, on net9.0 and net10.0 - the four service location statements emitted as `delta, charlie, bravo, alpha` in one run and `bravo, alpha, delta, charlie` in the next, everything else identical. Both pass with commit 2. At commit 1 `CodegenTests` is 468 total, 466 passed, 2 failed: the new ones and nothing else.

Full suite green otherwise - CodegenTests 468, CoreTests 602, CommandLineTests 295, EventTests 1141, EventStoreTests 143, plus the generator and Aspire suites, across net9.0 and net10.0. One unrelated flake: `EventTests.Daemon.BlueGreenSideEffectGateTests.fresh_deploy_suppresses_side_effects_to_the_prior_version_mark_then_enables_them` fails intermittently on net9.0 when the whole assembly runs. It reproduces on unmodified `main` - 9 of 12 runs there against 7 of 12 here - and nothing under `src/EventTests` touches code generation, so it is not fallout from this. Happy to open it separately if it is news.


jeremydmiller (approved):
Verified independently and this is right. Approving — with one note about CI that is not yours to fix.

## The load-bearing claim checks out

The fix rests on "the cache's keys are exactly the frames passed in", and it is worth saying precisely why that holds, because it is the whole safety argument:

- `DependencyGatherer`'s constructor ends with `foreach (var frame in frames) Dependencies.FillDefault(frame);`, and `FillDefault` is `Fill(key, _onMissing)` — so one entry per frame, with the **computed** dependency list as the value, not a null placeholder.
- Nothing between that and step 3 adds a key. `findInjectedFields` and `findSetters` read `Variables`, and `Variables.OnMissing` does not touch `Dependencies`.
- Step 3 is the first thing that mutates `frames`.

So at the moment step 3 runs, `Keys(Dependencies) == frames` exactly, and enumerating the cache's values is the same multiset as `frames.Select(f => cache[f])`. Set gathered is unchanged; only the order becomes deterministic. And since every `x` is already a key, `Dependencies[x]` never fires `OnMissing` — no extra walk, no new entries.

## `ToArray()` really is load-bearing, and for the reason you give

The enumeration source is now `frames`, and the loop mutates `frames` via `Insert(0, …)` — lazy would throw `InvalidOperationException`. The old form was safe only by accident of enumerating a collection nobody was touching. Worth the sentence you gave it in the PR body; a future reader will otherwise "simplify" it back.

One more equivalence that is easy to miss and holds: moving `Where(x => !frames.Contains(x))` from lazy to eager does **not** change the result. In the old code a `Contains` check could see frames inserted earlier in the same loop; in the new one every check happens before any insertion. Because `Distinct()` and `Contains` use the same (reference) equality, no surviving element can be made `Contains`-true by inserting a different one — so the two orderings of the checks agree. Insert-at-0 semantics, and therefore the reversed prefix, are unchanged either way.

Agreed on rejecting `OrderBy`: it is a stable sort, so with no stable key on `Frame` the ties fall straight back to hash order and you would have fixed nothing.

## The tests

I bisected rather than taking the numbers on trust:

| | CodegenTests (net10.0) |
|---|---|
| `0ddcd9e` (repro only) | **2 failed**, 466 passed, 468 total |
| `0c8a888` (with fix) | **468 passed** |

Exactly what you reported. The pair is the right shape too — the 26-iteration one describes the symptom an operator sees, and the hash-pinned one makes the failure deterministic so this cannot rot into a flaky test that gets muted. `generated_source_does_not_depend_on_frame_hash_codes` is the one I would keep if forced to choose.

The provenance in #832 — 15/60 files differing at 1.24.1, 6/60 at 1.31.0 after #196 — is what makes the diagnosis convincing rather than plausible, and it correctly identifies this as the remainder of the same bug rather than a new one. It also closes out JasperFx/wolverine#3059, which guessed at this without a repro.

## Two notes, neither blocking

**CI has not run.** The check is `action_required` — GitHub gates workflow runs on fork PRs from outside contributors until a maintainer approves them. Nothing wrong with your branch. I have read the whole diff (one new test file plus a one-statement source change; no workflow files, no build scripts) and run the suite locally, so the approval is a formality, but a maintainer needs to click it before the merge has a green check against it.

**The `BlueGreenSideEffectGateTests` flake is known and is not yours** — it reproduces on unmodified `main`, as you found. No need to open anything; thanks for checking rather than assuming.

Nice piece of work: the diagnosis, the provenance, the deterministic repro, and a one-statement fix that does not widen the surface.

PR #838, opened 2026-09-13
gh-832: order step 3's missing-frame insertion by the method's own frame list (1.0 backport)

AllainPL:
Backport of #833 to the 1.0 branch. Straight cherry-pick of both commits, nothing adapted.

Thanks for turning the original round so quickly and cutting 2.69.2 the same day.

## Why

Step 3 of `MethodFrameArranger.compileFrames` on this branch is byte for byte the walk that was fixed on main: `DependencyGatherer.Dependencies` is an `ImHashMap` keyed by `Frame`, `Frame` does not override `GetHashCode`, so the walk follows identity hash codes and the statements inside generated method bodies reorder between runs.

2.69.2 carries the fix on 2.x. Wolverine 5 pins JasperFx 1.x, so anything still on that line does not pick it up through #4421 - which is where we are.

## What is here

The two commits from #833, unchanged:

- the two failing tests
- the one-statement fix

The test file compiles here as-is. It uses only `[Fact]` and Shouldly, which behave the same under xunit 2.9.3 as under xunit.v3, so no port was needed. `CodeGeneration_ordering_determinism` from #196 is already on this branch, and the new class sits alongside it.

## Test plan

- Without the fix, the two new tests fail 5 runs out of 5.
- With it, `CodegenTests` is 323 passed, 0 failed, across net8.0, net9.0 and net10.0.

If the 1.x line is not shipping any more, close this - no argument from me.


jeremydmiller:
@AllainPL Hey man, I'm really not enthusiastic about the back port. Is there some reason why you can't upgrade? We generally aren't back porting if you're not a customer.

AllainPL:
That's fair, and I'd rather not have you carrying a branch on our account.

For what it's worth, the thing that gave us pause on 6 was `ServiceLocationPolicy` defaulting to `NotAllowed` - 65 of our 66 generated handlers service-locate, so at a glance it looked like a much bigger job than it is. `RestoreV5Defaults` looks like the way through and we'll probably take it.

Closing this. Thanks again for the quick turnaround on main.

jeremydmiller:
"65 of our 66 generated handlers service-locate" -- why?

AllainPL:
Inherited codebase we're still modernizing, so plenty of it is ours to clean up.

It's two roots rather than 65 decisions though. One is a forwarding registration - a scoped concrete with two interfaces aliased onto it via `sp => sp.GetRequiredService<T>()`, so all three are the same instance per scope. Middleware writes through one interface and handlers read through the other, so it has to be one object, and MS DI has no first-class way to alias without the lambda.

The other is an EF DbContext through `AddDbContext`'s options lambda, and that one spreads - anything whose constructor takes it falls back too. We're on the built-in container, no Lamar.

Since you're asking though - is either of those inlinable in a way we're missing? If a DbContext just isn't, then 65 of 66 is normal for an EF-backed app rather than a smell, which is worth knowing before we reach for `RestoreV5Defaults`.

Issue #743, opened 2026-09-04
Codegen: let an ICodeFile emit type/method attributes and know the target language, so generated code can carry its own Native AOT roots


jeremydmiller:
Follow-up from JasperFx/wolverine#4287. The missing piece of Wolverine's Native AOT startup story is that the artifacts \`codegen write\` produces — the \`GeneratedHandlerRegistry\` and the pre-built \`MessageHandler\` types — are only ever reached via reflection (\`Assembly.ExportedTypes\` + \`Activator.CreateInstance\`), so ILC trims them and \`TypeLoadMode.Static\` silently falls back to an assembly scan that finds nothing under AOT. Today the app author has to hand-write \`[DynamicDependency(All, ...)]\` roots for the registry, each generated handler type, each handler class, and the closed \`MessageRouter<T>\`/\`EmptyMessageRouter<T>\` per message type (verified working end-to-end in wolverine#4287).

The obvious owner for those roots is the generated code itself: \`codegen write\` knows every one of those types, so the emitted registry file could carry

- a \`[DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(...))]\` block on the registry class for each handler type, message type, generated handler type, and closed router generic, and
- a tiny \`[ModuleInitializer]\`-rooted companion class referencing the registry, so the whole graph is anchored with zero app-side code (module initializers are unconditional ILC roots).

Two small codegen capabilities are missing to build that cleanly in Wolverine:

1. **Attribute emission.** \`GeneratedType.Header\` / \`GeneratedMethod.Header\` (\`ICodeFragment\`) can technically smuggle attribute text in front of a declaration, but there is no first-class way to say "this generated type/method carries these attributes", and nothing that renders them per-language.
2. **Language awareness.** \`DynamicCodeBuilder.Language\` (csharp/fsharp) never reaches \`ICodeFile.AssembleTypes\` or the fragments, so a C#-attribute Header would corrupt \`codegen write --language fsharp\` output. Either surface the language to the code files/fragments, or make attribute emission a modeled concept that each language writer renders correctly.

With those two in place, Wolverine can make \`TypeLoadMode.Static\` + \`PublishAot\` work with no hand-written rooting at all. Happy to do the Wolverine side once the codegen surface exists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jeremydmiller:
[...]

Issue #227, opened 2026-05-11
Codegen CLI commands (preview / write / test) bypass ServiceLocationPolicy enforcement


jeremydmiller:
Parking note from downstream consumer (Wolverine), to address as part of 2.0.

## Symptom

Downstream consumers (Wolverine, Marten) call back into a chain-level
`AssertServiceLocationsAreAllowed(reports, services)` hook to enforce
their `ServiceLocationPolicy` (`AllowedButWarn` / `AlwaysAllowed` /
`NotAllowed`) when JasperFx codegen would otherwise emit a service-
location call (lambda-factory registration, opaque to codegen).

The hook is correctly invoked during the runtime compile path
(`DynamicCodeBuilder.TryBuildAndCompileAll` / cache-miss
compilation). It is **not** invoked from the CLI codegen paths:

- `dotnet run -- codegen preview` → `GenerateCode(svs)` on a
  single chain assembly
- `dotnet run -- codegen write` → `DynamicCodeBuilder.GenerateAllCode`
- `dotnet run -- codegen test`

A host that sets `ServiceLocationPolicy.NotAllowed` gets the
expected `InvalidServiceLocationException` at runtime but a clean
"all generated" result from any of the three CLI commands.

Reported against Wolverine in https://github.com/JasperFx/wolverine/issues/2718.

## Reproducer

```csharp
builder.Services.AddScoped<IGlobalUserContext>(sp => new GlobalUserContext());
builder.Host.UseWolverine(opts =>
{
    opts.CodeGeneration.TypeLoadMode = TypeLoadMode.Auto;
    opts.ServiceLocationPolicy = ServiceLocationPolicy.NotAllowed;
});

// Endpoint using the lambda-scoped service:
public static class Endpoints
{
    [WolverineGet("/ping")]
    public static async Task<IResult> Ping(IGlobalUserContext ctx)
        => Results.Ok();
}

// dotnet run -- codegen preview
// → emits GetRequiredService<IGlobalUserContext> instead of throwing
```

## Proposed direction (JasperFx 2.0)

Add the same `AssertServiceLocationsAreAllowed` callback invocation
to `GenerateCode(svs)` and `GenerateAllCode(...)` so the three CLI
paths enforce policy consistently with the runtime compile path.

Concretely the two surfaces that need to call back into the chain
hook are inside `GeneratedAssembly.GenerateCode(IServiceVariableSource?)`
[...]

PR #867, opened 2026-09-18
wolverine#4486: `codegen test` compiles a file whose AotRoots name a sibling generated type

### LINKED ISSUE #4486: codegen test fails with CS0234 for every message handler since 6.37.0 (AotRoots references sibling types across per-file compilations)
## Summary

Since **6.37.0** (#4426 / 77349ebc), `dotnet run -- codegen test` fails on a clean checkout (no pre-generated code in the application assembly) for any application that has **message handlers**. Every message handler shows up as a `CS0234`:

```
JasperFx.CodeGeneration.GeneratorCompilationFailureException: Failure when
trying to generate code for Wolverine.Runtime.Handlers.HandlerGraph
     System.InvalidOperationException: Compilation failures!

     CS0234: The type or namespace name 'AgendaAlteradaHandler2048194527' does
     not exist in the namespace 'Internal.Generated.WolverineHandlers' (are you
     missing an assembly reference?)
     CS0234: The type or namespace name 'ExpurgarAtendimentosHandler1558524863'
     does not exist in the namespace 'Internal.Generated.WolverineHandlers' ...
     (one per message handler; HTTP endpoints are unaffected)
```

`codegen write` followed by `dotnet build` compiles fine, and so does the app in `TypeLoadMode.Static`. Only `codegen test` fails, which is what we (and probably others) use as the PR gate for pre-generated code.

## Versions (bisected)

| WolverineFx | `codegen test` |
|---|---|
| 6.30.3 | passes |
| 6.35.0 | passes |
| 6.36.0 | passes |
| 6.37.0 | **fails** |
| 6.38.0 | **fails** |
| 6.39.0 | **fails** (JasperFx 2.73.1) |

.NET 10, Marten 9.x, Wolverine.Http + Wolverine.Kafka + local queues. It's the same result in four independent services.

## Root cause

Two changes collide:

1. **JasperFx #227**: `DynamicCodeBuilder.TryBuildAndCompileAll` (the `codegen test` path) now compiles **each `ICodeFile` into its own `GeneratedAssembly`** (`collection.StartAssembly(...)` inside `foreach (var file in collection.BuildFiles())`), so `ServiceLocationPolicy` can be enforced per file.
2. **Wolverine #4426**: `HandlerGraph.explodeAllFiles` collects `generatedHandlerTypeNames`, and `HandlerRegistryCodeFile` feeds them to `GeneratedAssembly.AddAotRoots`. That emits `[DynamicDependency(..., typeof(global::Internal.Generated.WolverineHandlers.<HandlerTypeName>))]` for every sibling generated handler, by **name**.

Under `codegen test`, the `GeneratedHandlerRegistry` file compiles in isolation, so the handler types that `AotRoots` names by string live in *other* in-memory assemblies, and every one fails with `CS0234`. HTTP-only apps don't hit this, because their `AotRoots` roots only the registry itself, which is in the same file. That matches the "HTTP-only projects get a single root" note in the #4430 description.

Two supporting observations:
- If `codegen write` + `dotnet build` ran first (generated types already in the app assembly), `codegen test` passes. The in-memory compile then resolves the names against the application assembly. In that situation it no longer tests freshly generated code.
- The drift gate (#4427) uses `codegen write`, and the AotSmoke projects use `publish`. Neither path runs `codegen test` against a project with message handlers, so CI didn't catch this.

## Repro

Any app with at least one conventional message handler:

```bash
rm -rf Internal/Generated
dotnet build
dotnet run -- codegen test   # CS0234 for each message handler
```

## Possible fixes

- Skip `AddAotRoots` when not writing to disk (`codegen test` / dynamic compile). The roots only matter for the committed file that ILC sees.
- Or compile the registry file together with the handler files in `TryBuildAndCompileAll`.
- Or emit the roots in a separate `ICodeFile` that is excluded from the per-file test compile.

Workaround for now: pin WolverineFx 6.36.0.

andrevlins:
Fixes JasperFx/wolverine#4486.

## Problem

`codegen test` (`DynamicCodeBuilder.TryBuildAndCompileAll`) compiles every `ICodeFile` into its own in-memory assembly (#227). If an `AotRoots` companion roots a sibling generated type through `AttributeArg.TypeNamed` (#743), it references a type that lives in a different assembly, and the compile fails with `CS0234`.

Since Wolverine 6.37 (JasperFx/wolverine#4426), `HandlerRegistryCodeFile` roots every generated message handler this way. So `codegen test` now fails for any app with message handlers, while `codegen write` + build works fine.

## Fix

- `TryBuildAndCompileAll` sets a new `GeneratedAssembly.CompiledInIsolation` flag before `AssembleTypes`, next to where it already sets `TargetLanguage`.
- When the flag is set, `AddAotRoots` drops the `TypeNamed` roots that don't resolve to a type of the same assembly. Roots to existing `Type`s and to the file's own generated types are kept.
- Only `TryBuildAndCompileAll` sets the flag, because its code is never written to disk. `codegen write` and the Auto-mode `CompileAndAttach` (which writes what it compiles) are unchanged and still emit every root into the file the trimmer sees.

Scope is limited to `AddAotRoots`. A hand-built `GeneratedAttribute` that names a sibling through `TypeNamed` would still fail under `codegen test`. I left that alone because `AddAotRoots` is the documented way to emit rooting.

## Tests

- New `AotRootsIsolatedCompilationTests`, with two files where one roots the other by name:
  - `TryBuildAndCompileAll` compiles. Before the fix this reproduced the exact `CS0234`.
  - In the isolated compile, the own-type and existing-type roots stay and the sibling root is dropped.
  - `GenerateAllCode` (the write path) still roots every sibling.
  - A standalone `GeneratedAssembly` is not filtered.
- CodegenTests 472/472, CoreTests 607/608 (1 pre-existing skip), CommandLineTests 295/295, EventTests 1258/1258, CodegenTests.FSharp 1/1. All on both net9.0 and net10.0.
- `docs/codegen/aot.md` has a new note.

## Verified against a real app

Four services on Wolverine 6.39.0 / Marten 9.37.0, using this patch applied on top of `V2.73.1` (the version Wolverine 6.39 references) and packed locally:

| | stock JasperFx 2.73.1 | patched |
|---|---|---|
| `codegen test`, clean tree | fails, 19 × `CS0234` | passes in all 4 services |
| `codegen write` output | 63 files | **byte-identical** (`diff -r`), all 20 generated-type roots present |
| build with the written code (Static) | builds | builds |

Once this ships, Wolverine only needs to pick up the JasperFx version; no Wolverine change is required. It would also be worth running `codegen test` in Wolverine's CI against a project that has message handlers (e.g. `src/Testing/ConsoleApp`). Neither the drift gate (`codegen write`) nor the AotSmoke projects (`publish`) exercise this path.


jeremydmiller:
@andrevlins Thank you! The AOT "compliance" has turned out to be an ongoing challenge:(

